PR #2, opened 2015-08-17
Add missing requirements

praseodym:


jmgrosen:
Amazing to see people are actually trying to use this! :)

I'm afraid you'll probably be pretty disappointed with the current state, but we're hard at work on it (well, or will be shortly)!

praseodym:
Cheers! I'm looking forward to seeing what's in store.

PR #3, opened 2015-08-20
Added PySide and pygraphviz as requirement

riyadparvez:


praseodym:
`qtconsole` is already included in `ipython>=3.0.0,<4.0`. When upgrading to `ipython>=4.0.0` it is needed, however (it got moved out of the regular `ipython` package).

I had installed `python-pyside`, `python-graphviz` and `qt4-dev-tools` through apt-get (Debian jessie) because I had pip build issues with those.

riyadparvez:
I also have `ipython` installed, but didn't have `pyside` installed as a requirement of `ipython`.

jmgrosen:
Hmmm... I haven't actually tested it with PySide, only PyQt4. Does it work alright?

Pygraphviz should definitely be added, though.

praseodym:
PySide works fine for me.

PR #31, opened 2017-06-22
Add script to show GUI from Docker container on Mac

tiffanyb:
Turns out that Mac turns off the 6000 port for listening to the connection.


ltfish:
Can you please rename it to `run-docker-macos.sh`?

tiffanyb:
@ltfish would you like to also rename run-docker-linux.sh?

ltfish:
Thank you!

PR #36, opened 2017-10-27
Adds instructions for fixing MIPS_GRP_CALL issue

Jake-R:
The instructions on fixing the MIPS_GRP_CALL issue are not clear. The angr fork of capstone does not have the correct commit included in master and neither does the latest mainline capstone. In order to install it you must change branches to 'next' which does have it included.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,17 @@ python start.py
 
 ### 'module' object has noattribute 'MIPS_GRP_CALL'
 
-Your capstone install is too old.
-You may install the version from [our fork](https://github.com/angr/capstone).
+Your capstone install does not support functionality that angr-management uses.
+
+To install a version that does:
+```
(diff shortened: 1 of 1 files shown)


zardus:
Could you change these directions to work with a virtualenv rather than using sudo?

Jake-R:
I've added another commit that isolates the bindings and libcapstone4 to the current virtualenv. It also isolates libcapstone4 to just the python bindings so tools like r2 with still load capstone 3.x.

ltfish:
> The angr fork of capstone does not have the correct commit included in master and neither does the latest mainline capstone.

It was an ignorance on our end - someone updated angr/capstone and forgot to use the `next` branch instead of `master`.

Thanks for providing the correct solution!

PR #47, opened 2018-07-26
Refactor synchronization mechanism

rhelmot:
So I was trying to do something totally different and I was really fed up with how cumbersome it was to have two components being synchronized, so I did this.

There is a new class, ObjectContainer, which acts as a transparent proxy for the object it holds. If you have `simgr_container = ObjectContainer(simgr)`, you can access `simgr_container.active[0]` and stuff. It does three important things:

1. Holds a reference to an object, so that if multiple objects hold the container, their references to the contained object will all be updated automatically
2. Lets you give objects human-readable names
3. Adds a signaling API to trigger callbacks on holders of the container

The last thing is important. You can subscribe to a container's events with `container.am_subscribe(callback)` (all container attributes are prefixed with `am_` to avoid conflicts with the held object), and then unsubscribe with `container.am_unsubscribe`. You issue an event with `container.am_event(**kwargs)`, and then all subscribers will get `callback(**kwargs)` called.

This is a very ad-hoc system, but the intuition is that by far the most common case is when the object has been updated in some way and the holders need to update their views based on it. They don't care what exactly happened, they just know _something_ happened, so they ignore kwargs and do their updates. From there, nuance can be added by looking at any arguments you think might be important.

A lot of things got vastly simplified as a result of this - SimgrDescriptor is no more, SimulationManagers is no more (it's just a container of a list), and starting jobs can use an alternate constructor which will just send a signal to the analysis subject when it's finished. And now, you can click haphazardly between different simgrs and states in the symbolic execution view, and everything will get highlighted and refreshed immediately!

This will be a good way to implement something like IDA's "keep in sync with this view" feature - different views can expose a public API which has different ObjectContainers available, and by making a reference to that container, you will be kept in sync.

Three todos/regressions:

1. I've just been hapahazardly using pylint's autocomplete, which generates import patterns which don't seem very pretty, those should be fixed up at some point
2. We need to add events to each view for when it's closed, so that all its components can unsubscribe from their objects, otherwise they will continue to receive events after they've been closed and all hell will break loose
3. There seem to be segfaults happening nondeterministically during application startup? I don't know if this was happening before but it's probably my fault. It's easy to get into infinite recursion loops if you use these APIs wrong, but I have no idea how that could be nondeterministic.

### DIFF
--- a/angrmanagement/data/instance.py
+++ b/angrmanagement/data/instance.py
@@ -4,7 +4,7 @@ from Queue import Queue
 
 import ana
 
-from angrmanagement.data.states import StateRecord
+from .states import StateRecord
 from .jobs import CFGGenerationJob
 from ..logic import GlobalInfo
 from ..logic.threads import gui_thread_schedule_async
--- a/angrmanagement/ui/dialogs/new_path.py
+++ b/angrmanagement/ui/dialogs/new_path.py
@@ -3,10 +3,10 @@ from PySide.QtGui import QDialog, QVBoxLayout, QHBoxLayout, QLabel, QPushButton,
 from PySide.QtCore import Qt
 
 from angr import StateHierarchy
-from angrmanagement.data.instance import ObjectContainer
 
-from ..widgets import QAddressInput, QStateComboBox
+from ...data.instance import ObjectContainer
 from ...utils.namegen import NameGenerator
+from ..widgets import QAddressInput, QStateComboBox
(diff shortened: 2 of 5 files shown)


rhelmot:
Oh right, another thing, this also creates the instance before the workspace, since the workspace needs the instance's object containers, such as the list of simgrs, to propogate down into the views.

PR #53, opened 2018-12-25
Support displaying results during the initial analysis phase

ltfish:
- You can now view functions and CFG before the entire CFG is generated.
- GUI is slightly more responsive during the initial analysis phase.
- The linear disassembly viewer is now generated progressively. We no longer pre-generate all blocks, instructions, and operands of the entire binary upon binary loading.


rhelmot:
Holy moly, merry Christmas

ltfish:
Merry Christmas!

PR #63, opened 2019-03-21
Add comment support

sraboy:
This allows adding, editing, and displaying comments from the KB. Like IDA, comments take precedence over any strings in the display.

comment

Some things I'd like input on:
 - Should I keep the text blue (like IDA)?
 - Should I use a wider spacing following operands (currently uses `5`, like strings)
 - Should I prepend a semicolon in the display?

Also, there is a lot of duplication in `QInstruction`'s linear/graph painting functions. I followed along with that for now but I plan to refactor those out.


sraboy:
Not sure if the KB interface is going to change anytime soon but this is good to go.

ltfish:
Thank you! Just merged.

We can discuss those things later.

PR #65, opened 2019-03-22
Add Function/Instruction Selection Events

sraboy:
I'm adding various signals to the "view" objects. I've got a bunch of hacks doing this right now but I'd like to add standard signals as well as some callbacks for things like highlighting instructions or rows in the functions view.

Right now, this PR is for code review. I've got about a dozen signals/callbacks to add in to at least get through my current project.

### DIFF
--- a/angrmanagement/ui/views/disassembly_view.py
+++ b/angrmanagement/ui/views/disassembly_view.py
@@ -2,7 +2,7 @@
 from PySide2.QtWidgets import QVBoxLayout, QMenu, QApplication
 from PySide2.QtCore import Qt, QSize
 
-from angrmanagement.data.instance import ObjectContainer
+from ...data.instance import ObjectContainer
 from ...utils import locate_function
 from ...data.function_graph import FunctionGraph
 from ...logic.disassembly import JumpHistory, InfoDock
--- a/angrmanagement/ui/views/disassembly_view.py
+++ b/angrmanagement/ui/views/disassembly_view.py
@@ -1,6 +1,6 @@
 
 from PySide2.QtWidgets import QVBoxLayout, QMenu, QApplication
-from PySide2.QtCore import Qt, QSize
+from PySide2.QtCore import Qt, QSize, Signal
 
 from ...utils import locate_function
 from ...data.function_graph import FunctionGraph
@@ -15,6 +15,8 @@ from .view import BaseView
 
 
 class DisassemblyView(BaseView):
+    sig_insn_selected = Signal(int)
+
     def __init__(self, workspace, *args, **kwargs):
         super(DisassemblyView, self).__init__('disassembly', workspace, *args, **kwargs)
 
@@ -240,6 +242,7 @@ class DisassemblyView(BaseView):
         else:
             self.current_graph.select_instruction(insn_addr, unique=QApplication.keyboardModifiers() & Qt.CTRL == 0)
             self.current_graph.show_instruction(insn_addr)
+            self.sig_insn_selected.emit(insn_addr)
 
     def toggle_operand_selection(self, insn_addr, operand_idx):
         """
--- a/angrmanagement/ui/views/disassembly_view.py
+++ b/angrmanagement/ui/views/disassembly_view.py
@@ -154,7 +154,7 @@ class DisassemblyView(BaseView):
     # Public methods
     #
 
-    def set_cb_on_insn_select(self, callback):
+    def subscribe_insn_select(self, callback):
         self._linear_viewer.selected_insns.am_subscribe(callback)
         self._flow_graph.selected_insns.am_subscribe(callback)
 
--- a/angrmanagement/ui/views/disassembly_view.py
+++ b/angrmanagement/ui/views/disassembly_view.py
@@ -155,6 +155,14 @@ class DisassemblyView(BaseView):
     #
 
     def subscribe_insn_select(self, callback):
+        """
+        Appends the provided function to the list of callbacks to be called when an instruction is selected in the
+        disassembly. The callback's parameters are:
+            'graph': the `QBaseGraph` object
+            'addr': integer address of the selected instruction
+            'block': the `QBlock` containing the instruction
+        :param callback: The callback function to call, which must accept **kwargs
+        """
         self._linear_viewer.selected_insns.am_subscribe(callback)
         self._flow_graph.selected_insns.am_subscribe(callback)
 
--- a/angrmanagement/ui/views/disassembly_view.py
+++ b/angrmanagement/ui/views/disassembly_view.py
@@ -1,6 +1,6 @@
 
 from PySide2.QtWidgets import QVBoxLayout, QMenu, QApplication
-from PySide2.QtCore import Qt, QSize, Signal
+from PySide2.QtCore import Qt, QSize
 
 from ...utils import locate_function
 from ...data.function_graph import FunctionGraph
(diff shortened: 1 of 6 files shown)


rhelmot:
Interestingly enough, we have a homebrew synchronization mechanism that's designed to work in our data layer rather than in the qt layer: https://github.com/angr/angr-management/pull/47

Is ObjectContainer or Signal more useful to implement the kinds of communication you're looking for?

rhelmot:
Actually yes, I'm gonna say that ObjectContainer is definitely the better tool for the job here, since the things you've tweaked so far are all of the same form as the stuff objectcontainer was for: synchronizing which data element the user currently has focused.

sraboy:
To be honest, I just wasn't entirely sure what was going on with the simgr code. I can switch to `ObjectContainer` once I understand how to use it. Does each view object need to get wrapped in its own container or something?

rhelmot:
ObjectContainer is used when there is one common "reference" that needs to be updated between many components and possibly exposed to the user through the embedded terminal. There is exactly one ObjectContainer per synchronization point, and everyone who wants to synchronize it holds a reference to that ObjectContainer. That ObjectContainer holds a reference to the current object that is focused or shared or whatever, and this reference changes.

When you want to update it, you can do so and then fire an event which will call callbacks on everyone registered to indicate that they might want to update their display accordingly.

sraboy:
Sounds good. I'll start tinkering and port these two Signals to ObjectContainers and push here again.

sraboy:
@rhelmot What do you think of the current implementation? You can apply this patch to `main_window.py` for a quick test:

```
diff --git a/angrmanagement/ui/main_window.py b/angrmanagement/ui/main_window.py
index b60f024..1a23d0a 100644
--- a/angrmanagement/ui/main_window.py
+++ b/angrmanagement/ui/main_window.py
@@ -18,6 +18,17 @@ from .dialogs.load_binary import LoadBinary, LoadBinaryError
 from .dialogs.new_state import NewState
 from .toolbars import StatesToolbar, AnalysisToolbar, FileToolbar
 
+def on_insn_select(**kwargs):
+    addr = kwargs.get('addr', None)
+    if addr is not None:
+        print("User selected {:#010x}".format(addr))
+    else:
+        print("Why are we here?")
+
+def on_func_select(**kwargs):
+    func = kwargs.get('func', None)
+    if func:
+        print("User selected function '{}' at {:#010x}".format(func.name, func.addr))
 
 class MainWindow(QMainWindow):
     """
@@ -183,6 +194,8 @@ class MainWindow(QMainWindow):
         right_dockable_views[0].raise_()
 
         self.central_widget.setTabPosition(Qt.RightDockWidgetArea, QTabWidget.North)
+        self.workspace.views_by_category['disassembly'][0].subscribe_insn_select(on_insn_select)
+        self.workspace.views_by_category['functions'][0].subscribe_func_select(on_func_select)
 
     #
     # Event
```

If you like it, you can feel free to merge now. I have more callbacks to add but may need some time.

angrmanagement/ui/views/disassembly_view.py
```diff
@@ -2,6 +2,7 @@
 from PySide2.QtWidgets import QVBoxLayout, QMenu, QApplication
 from PySide2.QtCore import Qt, QSize
 
+from angrmanagement.data.instance import ObjectContainer
```
rhelmot: prefer relative imports

angrmanagement/ui/views/disassembly_view.py
```diff
@@ -153,6 +154,18 @@ def popup_xref_dialog(self, variable, asynch=True):
...
+        """
+        Appends the provided function to the list of callbacks to be called when an instruction is selected in the
+        disassembly. The callback's parameters are:
+            'graph': the `QBaseGraph` object
+            'addr': integer address of the selected instruction
+            'block': the `QBlock` containing the instruction
+        :param callback: The callback function to call, which must accept **kwargs
+        """
+        self._linear_viewer.selected_insns.am_subscribe(callback)
+        self._flow_graph.selected_insns.am_subscribe(callback)
```
rhelmot: Should these just be the same object? It would be really nice if the linear disassembly and the disassembly graph had the same selections when you toggle them

angrmanagement/ui/views/disassembly_view.py
rhelmot: here and several places below

sraboy:
I agree this is a bit of a wonky implementation but the layers/architecture isn't immediately obvious to me. Is the `Instance` meant to be like an MVP Presenter to which all the objects should be pushing their changes?

The current implementation (i.e., on master, not here) is done differently in these two cases:

---
User click-select is picked up by Qt in `QInstruction` which passes its addr to `DisassemblyView.toggle_instruction_selection()` which passes the addr it to `current_graph.select_instruction()` which passes the addr to the (non-derived) `QBaseGraph.select_instruction()`, which adds it to the `set()` of currently-selected instructions (only within the actual graph in view; i.e., `DisassemblyView.current_graph`).

How should that change? Since `QBaseGraph` contains the "data" of `selected_insns`, I wrapped that in `ObjectContainer`.

---
User click-select is picked up by Qt in `QFunctionTableView` which passes the func to `FunctionsView.selection_callback` which passes it to `Workspace.on_function_selected()` which passes it to `DisassemblyView.display_function` which also happens to update an internal `_selected_function`.

---

angrmanagement/ui/views/disassembly_view.py
sraboy: The current implementation has `DisassemblyView` with two separate `QBaseGraph` objects, each with their own `selected_insns: set()`.  On select, `DisassemblyView` uses its `current_graph` property to figure out which graph is being used and only toggles that graph's selection.

`selected_insns` could be pulled up into `DisassemblyView` (or the `Workspace.instance`).

angrmanagement/ui/views/disassembly_view.py
sraboy: Oops...

rhelmot:
agh. I'll deal with the refactors later myself.

tbh I don't understand MVC or MVP, but in my mind the instance is part of the data model? I consider the currently selective or active stuff to be data as opposed to presentation, but maybe that's wrong...

sraboy:
> tbh I don't understand MVC or MVP

I don't think anyone truly does... I can rationalize any bit being part of M, V, or P depending on how you look at it.

> in my mind the instance is part of the data model

Ah, okay. Well if you happen to get around to doing some of the refactoring during CHESS, I'll assist once I get a feel for how you want it all set up.

angrmanagement/ui/widgets/qfunction_table.py
```diff
@@ -8,6 +8,8 @@
 from PySide2.QtGui import QBrush, QColor
 from PySide2.QtCore import Qt, QSize, QAbstractTableModel, SIGNAL, QEvent
 
+from angrmanagement.data.instance import ObjectContainer
```
ltfish: I prefer a relative import (`from ....data.instance import ObjectContainer`). Why is absolute import better in this case?

sraboy: Because PyCharm does it automatically and then I forget to change it...

ltfish:
> Right now, this PR is for code review.

I was about to click "merge" before I noticed this line... Is this still WIP? Are you going to push more commits to this PR, or should I merge the code in as it is?

sraboy:
@ltfish I'm good to go with this if you are. The other changes will require a plugin interface so I'll do those in a separate PR.

ltfish:
Merged. Thanks!

PR #67, opened 2019-04-04
Plug It In!

sraboy:
I just discovered draft PRs so I figured I'd open this in case anyone has input. I'm focused on making things work for CHESS at the moment but I'd eventually like to not have a fork to maintain.

I've got separate code locally that I'm porting to this new framework little by little so the threading stuff is absolute bare bones at the moment.

### DIFF
--- a/angrmanagement/data/instance.py
+++ b/angrmanagement/data/instance.py
@@ -161,6 +161,23 @@ class Instance(object):
         self.jobs.append(job)
         self._jobs_queue.put(job)
 
+    def save(self, loc):
+        with open(loc, 'wb') as f:
+            pickled = pickle.dumps(self)
+            store = ana.get_dl()._state_store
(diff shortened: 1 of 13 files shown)


rhelmot:
one thing that jumps out at me: the `self.member: type = value` syntax was introduced in python 3.6, and we're still targeting python 3.5.

sraboy:
Oh darn, nice catch... PEP526. This syntax has a lot of advantages though. Any idea if you guys intend to bump the min version to 3.6 with the next release? If not, I'll change all these to comment style.

rhelmot:
When there's no longer an LTS Ubuntu that ships with 3.5

sraboy:
Yikes, got a couple years left then. I'll get all those fixed.

sraboy:
Once it's merged, I think I have a couple small fixes on another branch I'll put up in a new PR.

ltfish:
Merged. Thank you!

PR #71, opened 2019-04-10
[WIP] More Plugs

sraboy:
I ported 100% of our CHESS-related functionality to the new plugin. These are the few changes required. 

Other than a few little nitpicks, I wrapped the project and CFG in an `ObjectContainer` so the plugin can be notified of changes.


sraboy:
Oh, I also disabled `TestPlugin` by default. No reason to have some thread looping and sleeping forever...

ltfish:
LGTM. Can you please fix the conflict (since we merged in another PR earlier)?

sraboy:
Changed to WIP for the moment. I've got the conflicts settled but it turns out I need to resolve the how-to-load-a-plugin-at-runtime problem before this is worth merging anyway.

@twizmwazin Is that something you were working on (loading/enabling plugins at runtime)?

sraboy:
I keep running into situations where I need to keep a separate internal branch. I'll reopen after #80 is merged and I can re-rebase.

Issue #79, opened 2019-04-15
Feature: User-Specific Config/Settings


sraboy:
We should have a `~/.config/angr/angr-management` (exact path TBD). This would allow user installation of plugins without permissions issues and (eventually) saving user configuration information.

I could use this with the `PluginManager` but I also think it should be standardized across angr's packages... there are a lot of common/regularly-used settings or shortcuts I enable that I could stuff into something like a `.gdbinit`.

I can open a separate angr issue if it's something you think we could explore.

PR #80, opened 2019-04-15
Dynamic Plugin Loading

sraboy:
The diff makes it look like a total rewrite. Here are the key points:

- Load all plugins dynamically, eliminating `register_default` and the entire awkward concept of 'default plugins'. It's either installed or not and either enabled or not. 
- Plugins get notified via `on_disable` so some manner of "system/default plugin" can notify the user appropriately if they're about to break stuff.
- Allow loading (installing) plugins from other packages (see the README updates).
- The new "Plugins" menu on the main menu bar brings up a new dialog, allowing users to enable/disable plugins. 
- Plugins have three seconds to stop when asked (e.g., closing the window). After that, we murder them mercilessly.
-No other runtime options are exposed right now so plugins will have to implement their own extra menus/dialogs/callbacks as needed (e.g., to start a thread on a plugin that does not autostart).

There are several TODOs in the code but it works for me™. I'll update #71 after this is merged.

Sync: #79.


ltfish:
Is this still WIP?

sraboy:
There's always room for improvement but it works pretty well as-is. I say merge away if you're happy with it. I've got plenty of other little tweaks I'll push upstream later.

ltfish:
Merged. Thanks!

PR #84, opened 2019-04-21
Added missing dependency nclib

l19980623:
Hello, it appears that a required dependency `nclib` is not included in `requirements.txt` and `setup.py`. Attempting to run without will result in an error:
```
...
 File "...\angr-management\angrmanagement\ui\views\interaction_view.py", line 15, in <module>
    import nclib
ModuleNotFoundError: No module named 'nclib'
```
After manually installing `nclib` I was able to run it without issues. Thanks!

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,3 +6,4 @@ pyzmq
 shiboken2<=5.12.0
 PySide2<=5.12.0
 qtconsole
+nclib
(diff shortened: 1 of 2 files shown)


ltfish:
Merged. Thank you!

rhelmot:
fish... you shouldn't have merged this. nclib is an optional dependency, and archr depends on it anyway.

PR #88, opened 2019-05-07
Make paintEvent on graphs call the super method

lockshaw:
Without calling the super method, functionality such as setting the background brush with setBackgroundBrush breaks.

### DIFF
--- a/angrmanagement/ui/widgets/qdisasm_graph.py
+++ b/angrmanagement/ui/widgets/qdisasm_graph.py
@@ -162,7 +162,7 @@ class QDisasmGraph(QBaseGraph):
 
         self._update_size()
 
-    def paintEvent(self, event):
+    def _paintEvent(self, event):
         """
         Paint the graph.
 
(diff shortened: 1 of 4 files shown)


ltfish:
Good catch!

Can we do `_paint_event()` instead of `_paintEvent()`?

sraboy:
Since you're in there, could we maybe standardize the other painting methods too (i.e., eliminate `QLinearGraphicsView._paint_objects` and `QDisasmGraph._paint`)?

`QLinearGraphicsView._paintEvent` calls `setFont` then `_paint_objects`. `QDisasmGraph._paintEvent` calls `_paint` which calls `setFont`. `QSymExecGraph` just does everything in `_paintEvent`. Those extra methods are so small, I think they should be eliminated so everything looks like `QSymExecGraph`.

ltfish:
@sraboy Fine by me.

@lockshaw Do you want to make those changes, too? :)

lockshaw:
@sraboy They shouldn't all be in `_paint_event` (since all of them are atm ignoring the `event` parameter, and there are times we want to ask them to paint without an accompanying event), but I will convert them all over to how `QDisasmGraph` does it (`_paint_event` and `_paint`)

sraboy:
Oh yeah, good catch. Every little bit towards more consistency is great and I'm trying to give @ltfish a break from my giant refactors 😄

lockshaw:
I'm going to close this since I'm doing a large refactor of the graph painting which will render this fix useless.

PR #89, opened 2019-05-07
Add basic toml configuration file

lockshaw:


ltfish:
Hi I was thinking of this, and I think [toml](https://github.com/toml-lang/toml) is better than yaml as configuration files. Do you want to give it a try instead?

lockshaw:
@ltfish I just glanced at the libraries for parsing toml, and none seem to support adding custom types (pyyaml does). Getting to just write `!color 0xff,0xff,0xff` seems like a nice advantage to pyyaml, and that can actually be improved with a bit more work so you could write something like `Color(0xff, 0xff, 0xff)`. Thoughts?

sraboy:
+1 for toml being so much simpler. 

For that color example, you could do `color = "0xff,0xff,0xff"` or `color = [0xff,0xff,0xff]`. You could even do `color.red = 255 [...so on...]` or an inline table `color = {r=255, g=255, b=255}`.

lockshaw:
@sraboy Was hoping to avoid those, but in hindsight they're actually not bad at all. Updated to use toml!

angrmanagement/config/__init__.py
```diff
@@ -2,9 +2,16 @@
 import os
 
 from .config_manager import ConfigurationManager
+from xdg import BaseDirectory
```
ltfish: Please move this line to right below `import os`.

angrmanagement/config/config_manager.py
```diff
@@ -1,7 +1,31 @@
 
-from PySide2.QtGui import QFont, QFontMetricsF
+from PySide2.QtGui import QFont, QFontMetricsF, QColor
 
 from .config_entry import ConfigurationEntry as CE
+import toml
+import logging
```
ltfish: Same - move them to the top of the file.

In general, the imports should follow the following order:

```
Modules come with Python

Third-party modules

angr modules

Relative imports from the current package
```

angrmanagement/config/config_manager.py
```diff
@@ -1,7 +1,31 @@
...
+
+_l = logging.getLogger(__name__)
+
+def color_constructor(config_option, value):
+    if type(value) is str:
```
ltfish: If this line is not hot, I would prefer `isinstance()` instead.

ltfish:
Thanks! Please address my nitpicks.

ltfish:
Oh, one more thing, why are we ditching `requirements.txt`? We can make `setup.py` automatically get requirements from `requirements.txt` instead, which sounds like a better solution to me (so we can still support `pip install -r`).

angrmanagement/config/config_manager.py
sraboy: Also `isinstance(value, basestring)` would cover `unicode` as well.

rhelmot:
[this was discussed](https://github.com/angr/angr/pull/1241) a while ago and we decided that we want setup.py and not requirements.txt. generally, requirements.txt are for projects which are not python packages.

ltfish:
Thanks!

PR #90, opened 2019-05-08
Minor Tweaks

sraboy:
Most of the commit history is already merged but the way I rebased means git is confused. As you can see in the diff:

- In `Instance`, don't trigger a CFG change event with the `async` setter
- Add `get_` prefix since we're using methods rather than properties in plugins
- Use keyword args when calling callbacks so everyone can just use `**kwargs` to keep a stabler API


ltfish:
Merged. Thanks!

PR #100, opened 2019-06-25
Move angr-management's drawing code back to Qt's built-in 

lockshaw:


angrmanagement/utils/subscriptions.py
```diff
@@ -0,0 +1,58 @@
+import logging
```
rhelmot: IMO this is a bad pattern. I set up object containers the way I did in order to minimize the amount of messages that would happen - in that system you explicitly choose when to notify subscribers instead of there being implicit events.

This also just feels like a lot of magic being embedded into places where we really want to know exactly what's going on.

angrmanagement/ui/widgets/qsymexec_graph.py
```diff
@@ -112,90 +88,65 @@ def show_selected(self):
...
-            block = self._get_block_by_pos(pos)
-            if block is not None:
-                block.on_mouse_doubleclicked(btn, self._to_graph_pos(pos))
-                event.accept()
-                return
-
-        super(QSymExecGraph, self).mouseDoubleClickEvent(event)
-
-    def _on_keypressed_event(self, key_event):
+    # def mousePressEvent(self, event):
```
rhelmot: delete instead of commenting out. we have git history if we really need to see what used to be

angrmanagement/ui/widgets/qphivariable.py
```diff
@@ -90,22 +91,28 @@ def refresh(self):
...
 
-        # subvariables
-        self._subvar_names = [ ]
-        self._subvar_idents = []
-        for subvar in self.variables:
-            name = "Unk" if not subvar.name else subvar.name
-            self._subvar_names.append(name)
-            ident = "<%s>" % ("Unk" if not subvar.ident else subvar.ident)
-            self._subvar_idents.append(ident)
+#         # variable name
```
rhelmot: delete. see below

ltfish:
Eventually I think I restored or reimplemented all the lost functionalities. I will merge this PR to master within the next 24 hours if there is no objection.

ltfish:
Again, we really need test cases for angr management...

PR #102, opened 2019-07-09
filter out MenuSeparator

tiffanyb:


angrmanagement/ui/menus/menu.py
```diff
@@ -54,7 +54,9 @@ def __init__(self, caption, parent=None):
 
     def action_by_key(self, key):
         if not self._keyed_entries:
-            self._keyed_entries = dict((ent.key, ent) for ent in self.entries)
+            entries = [ent for ent in self.entries if
+                    isinstance(ent, MenuEntry)]
+            self._keyed_entries = dict((ent.key, ent) for ent in entries)
```
rhelmot: you can just put the filter in the existing comprehension. `dict((ent.key, ent) for ent in entries if isinstance(ent, MenuEntry))`

PR #108, opened 2019-07-28
Add a feature map view.

ltfish:
TODOs:

- [x] Sync the arrow position back to disassembly view.
- [x] Configurable colors.

Leave for later:

- [ ] Optimize speed for binaries with tons of blocks in memory.

Sync: PR #100 .


rhelmot:
This is like, the idea situation for ObjectContainer synchronization! You can tick the sync box and remove the clunky refresh calls by adding a "current cursor address" data element to the instance, and having anyone who wants to synchronize on it subscribe to it.

PR #110, opened 2019-08-30
Point development install to angr-dev

twizmwazin:
Previous method suggested installing angr-management in a virtualenv by itself, pulling deps from pypi. This works at the moment since there haven't been any huge changes since the last release, but this will not work long-term. angr-dev is the simplest solution.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,26 +29,7 @@ Please consider having a development install if you plan to use latest features/
 
 ### Development Install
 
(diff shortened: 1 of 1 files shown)


ltfish:
https://github.com/angr/angr-dev/blob/master/setup.sh#L483

I suppose this line can be removed now? PySide2 installs just fine in a virtualenv.

twizmwazin:
@ltfish It might install just fine, but pyside2 still doesn't support being run under pypy, no? Travis doesn't matter since that is going away soon anyways.

PR #115, opened 2019-10-17
adding demo splash screen

hippwndon:


angrmanagement/__main__.py
```diff
@@ -60,6 +74,11 @@ def main(filepath=None):
 
     MainWindow(file_to_open=filepath if filepath else sys.argv[1] if len(sys.argv) > 1 else None)
 
+    while timer.elapsed() < 2000 :
+        app.processEvents()
+    splash.close()
+    #splash finish
+
```
ltfish: It seems to me that this splash window only displays for a fixed amount of time (2 seconds). Can we make it disappear when `MainWindow` finish initialization?

ltfish:
Merged. Thanks!

PR #119, opened 2019-11-01
fixed splitter bug and modified is_split to boolean. close #98

### LINKED ISSUE #98: Odd Dock Behavior
Ping @ananay

Ref code: https://github.com/angr/angr-management/blob/3d6d502b87b58d2adcba6461606a338672421828/angrmanagement/ui/view_manager.py#L43-L48

I want to display my plugin on the far-right, i.e., right of `DisassemblyView`. If I set its view's `default_docking_position` to left/top/bottom, behavior is as expected. If I set it to `'right'`, then the above code adds it to the central widget as a tab, like `DisassemblyView`, etc. 

The only way to have it tabbed on the right is to set `default_docking_position` to an invalid value, which means its `dock_area` gets set to `Qt.RightDockWidgetArea` and the above code will follow the second branch instead.

Here's what I want, which I can only get by calling `BaseView` with `default_docking_position='middle'`:

angrmanagment_dock

This works but has a downside/bug. That barrier you see between `DisassemblyView` and my `ChessPlugin` is actually two barriers, between which the "Split View" feature puts tabs and which can be resized to display a large empty area.

Here's that awkward empty space:

empty_space

Here's a split view tab:

tab_split

I *think* the simplest solution is for `ViewManager` to expose a function to add a tab to the central area that's automatically split out to a specified size (e.g., formalize what I'm doing to get `Qt.RightDockWidgetArea` and adding the widget to the `MainWindow` rather than `central_widget`). However, the better solution might be improving support in `ViewManager` for nested & grouped docking (see Woboq's second example [here](https://woboq.com/blog/qdockwidget-changes-in-56.html)).

hippwndon:


ltfish:
Merged. Thanks!

PR #120, opened 2019-11-17
Feat/multi trace overlay

etrickel:
This feature adds the capability to overlay multiple traces onto the CFG in the disassembly view and the functions in the function list.  

While similar to the traceview feature, this capability is focused on giving the user visibility into multiple traces.  For example, the input json file might be generated by tracing all the inputs generated by AFL.  The resulting json file contains the combined addresses and hit counts after tracing the binary while using all the inputs.  

The idea is that this quickly shows branches missed during the fuzzing of the binary so that a user could then form inputs that follow the missed branch.


tiffanyb:
@etrickel could you rebase commits and merge to the traceview branch first? Let's have one branch with both multi-trace and single-trace. You can assign me as the reviewer and I will check and merge.

ltfish:
Merged in PR #122.

PR #105, opened 2019-07-15
Traceview

tiffanyb:

PR #122, opened 2019-11-26
Plugin refactor and traceview

### LINKED ISSUE #121: Custom Toolbar Buttons
While you're redoing the plugin system anyway... there's no way for a plugin to add a toolbar button. Right now, the `QToolBar` is created on the fly, added, and then returned but there's no way to access that returned object. I simply cached that object in the `Toolbar` class in my fork [(diff here)](https://github.com/angr/angr-management/compare/master...sraboy:chess_dev).

Unfortunately, I can't push that upstream... because rules and such.

rhelmot:
Supercedes #105 
Closes #121 

Seeking feedback from @ltfish and @sraboy 

Feature list:
- traceview
  - load a basic block trace through json and visualize it in the disassembly
  - click on an instruction to view it in the trace
  - click on a spot in the trace to navigate there, shift+up/down to step through the trace
- multitrace view
  - load a collection of traces for coverage analysis in a special json format
  - ctrl+right-click on a block to load a trace containing that block
  - basic blocks not covered will be colored red
  - adds basic block coverage as a function coloring metric and column in the functions view
- new plugin architecture
  - actually usable
  - plugins can be loaded from wherever on your filesystem
  - autoload folders and blacklists available in settings (or will be when we have a settings menu)
  - plugins resemble exploration techniques, providing many instrumentation points
    - set colors for instructions, basic blocks, functions
    - custom draw instrumentation for instructions, basic blocks
    - add whatever widgets you want to the disassembly view (:grimacing:, see below)
    - click handlers for instructions/blocks (:grimacing:, see below)
    - dynamically add elements to context menus for disassembly view items
    - statically add elements to the plugin menu and the file toolbar
    - whatever else anyone wants (background threads?)
  - ability to add custom data elements to the `instance` for other plugins to synchronize with
  - error handling
    - print errors during instrumentation to console
    - if errors happen during sensitive events, like drawing, plugin will be automatically deactivated

Minor outstanding todos which can be addressed later:
- traceview will lag if you click on an instruction represented too many times in the trace
- the "instrument disassembly view" mechanism is totally intractable in terms of allowing for multiple plugins to coexist. what realistically needs to happen is that we should let the plugin provide new views that can be opened and docked with whatever views we want. that way we can ensure resources are handled properly
- we should perhaps make ObjectContainer callbacks weak references so that it's not possible for there to be dangling references to plugins through that
- some of the click handers would be better implemented as subscriptions to global "selected instructions/blocks/whatever" data containers
- perhaps the sample plugin should be made to do something interesting as a demo of all the features


angrmanagement/data/instance.py
```diff
@@ -1,43 +1,45 @@
...
-        self.interactions = ObjectContainer([], name='Saved program interactions')
-        self.interaction_protocols = ObjectContainer([PlainTextProtocol], name='Available interaction protocols')
-        self.sync = SyncControl(self)
+        self.extra_containers = {}
+        self._container_defaults = {}
+
+        self.register_container('simgrs', lambda: [], List[angr.SimulationManager], 'Global simulation managers list')
+        self.register_container('states', lambda: [], List[angr.SimState], 'Global states list')
+        self.register_container('patches', lambda: None, None, 'Global patches update notifier') # dummy
+        self.register_container('cfg_container', lambda: None, Optional[angr.analyses.cfg.CFGBase], "The current CFG")
```
sraboy: Took me a minute to figure out the busted import. Need to sync on https://github.com/angr/angr/commit/2a3bd28e69ac13d1a9dd8fc762d2aebbc32cd62f#diff-d186f9f9ffc6122f3256e71152e3f0c3

angrmanagement/ui/main_window.py
```diff
@@ -130,7 +130,7 @@ def progress(self, v):
     #
 
     def _open_mainfile_dialog(self):
-        file_path, _ = QFileDialog.getOpenFileName(self, "Open a binary", ".",
+        file_path, _ = QFileDialog.getOpenFileName(self, "Open a binary", "",
```
sraboy: I can no longer get the dialog to show. Undoing this change doesn't fix it either but something broke for me. I don't have Qt debugging set up but I get no errors from Python... it just silently fails and returns an empty string to `file_path`.

Also, maybe something like `QDir.home().absolutePath()` as a default location?

sraboy: Derp. I always use my CHESS plugin so I haven't actually tried loading a file normally in forever. This is apparently due to my use of PyCharm via snap with a newer PySide2 package [(bug report)](https://bugreports.qt.io/browse/PYSIDE-912).

angrmanagement/ui/dialogs/load_plugins.py
```diff
@@ -98,19 +92,37 @@ def _init_widgets(self):
...
 
     def _on_cancel_clicked(self):
         self.close()
+
+    def _on_load_clicked(self):
+        file_path, _ = QFileDialog.getOpenFileName(
+            self, "Open a plugin (select __init__.py for packages)", "", "Python files (*.py)")
+        plugins = load_plugins_from_file(file_path)
+
+        if not plugins:
```
sraboy: This error is also shown after hitting "Cancel" on the Load Plugins dialog.

angrmanagement/config/config_manager.py
```diff
@@ -57,6 +57,9 @@ def color_constructor(config_option, value):
     CE('feature_map_color_unknown', QColor, QColor(0xa, 0xa, 0xa)),
     CE('feature_map_color_delimiter', QColor, QColor(0, 0, 0)),
     CE('feature_map_color_data', QColor, QColor(0xc0, 0xc0, 0xc0)),
+    # plugins
+    CE('plugin_search_path', str, '$BUILTIN_PLUGINS:~/.local/share/angr-management/plugins'),
```
sraboy: Could we make this `AM_BUILTIN_PLUGINS` or something slightly less generic? This literally collides with another project of mine.

angrmanagement/plugins/base_plugin.py
```diff
@@ -1,79 +1,29 @@
...
-
-        self._plugin_manager = plugin_manager
-        self._workspace = workspace
-        self._thread_should_run = False
-        _l.info("Loaded {}".format(self.__class__.__name__))
-
-    def register_callbacks(self):
-        pass
+class BasePlugin:
+    __i_hold_this_abstraction_token = True
```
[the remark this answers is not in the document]
rhelmot: It makes it easy to make more abstract plugin classes

angrmanagement/config/config_manager.py
rhelmot: fixed. that's embarassing...

angrmanagement/ui/dialogs/load_plugins.py
rhelmot: Huh. The Qt documentation implies that hitting cancel will return None but it instead returns an empty string. I added a check for falsey results.

rhelmot:
@ltfish do you want this squashed?

ltfish:
I decide to not squash commits since we couldn't keep the history and contribution log otherwise. Thanks!

PR #125, opened 2019-12-03
Add kwargs/insn_addr to instruction selection am_event

### LINKED ISSUE #124: QInstruction Click Handler Bugs
1) Typo: `handle_insn_click` vs `PluginManager.handle_click_insn`
2) Also, the conditionals here consume the click event if it's just a normal left-click, so the plugins are never called. I use every selection of an address to track user activity, telemetry-style, so I can't rely on `InfoDock.selected_insns` (which doesn't tell me which was just clicked, just all of which are currently selected; it also doesn't give me right clicks).

https://github.com/angr/angr-management/blob/be2ae29279f75b8c057fbad5af0d80d468852059/angrmanagement/ui/widgets/qinstruction.py#L68

sraboy:
Closes #124.


rhelmot:
hell yeah

PR #126, opened 2019-12-04
Add Support for MenuSeparator()

sraboy:
This just helps break up the noise when there's a bunch of extra entries. Use case:

```
def build_context_menu_insn(self, item):
    return [None,
            ('Mark Generic &POI', lambda ctxm: self.on_ctx_menu_tag_poi(ctxm.insn_addr)),
            ('&Annotate POI', lambda ctxm: self.on_ctx_menu_annotate_poi(ctxm.insn_addr)),
            ('Add &Task', lambda ctxm: self.on_ctx_menu_add_task(ctxm.insn_addr))]
```


rhelmot:
This is good, but I think the usecase you're describing would be better served by having the plugin manager automatically insert separators between different plugins' commands.

PR #144, opened 2020-03-03
Add console script

twizmwazin:
This PR adds a console script, so that when installed, the command `angr-management` will be available in the users `$PATH`. This necessitated moving logic from `start.py` into `__main__.py`, where it probably should have been anyway.

### DIFF
--- a/angrmanagement/__main__.py
+++ b/angrmanagement/__main__.py
@@ -3,6 +3,11 @@ import os
 import ctypes
 import threading
 import time
+import warnings
+
+def shut_up(*args, **kwargs):
+    return
+warnings.simplefilter = shut_up
 
 
 def check_dependencies():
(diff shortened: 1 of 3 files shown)


angrmanagement/__main__.py
```diff
@@ -3,25 +3,11 @@
...
-        return False
-
-    try:
-        import qtconsole
-    except ImportError:
-        sys.stderr.write("Cannot find the qtconsole package. You may install it via pip:\n" +
-                         "    pip install qtconsole\n")
-        return False
-
-    return True
```
ltfish: As discussed, let's replace it with a proper implementation using `distutils.core.run_setup`.

twizmwazin:
I've reverted my removal of check_dependencies, I'll make another PR since it is really a separate issue.

PR #146, opened 2020-03-12
Visualizing AFL bitmaps

Lukas-Dresel:
This adds the ability to display the coverage as recorded in AFL's bitmap. This only works for bitmaps created from QEMU mode as we do not know the compile time constants used by the source instrumentation. 

Branch collisions occur quite frequently in QEMU mode, so seeing single branches erroneously marked as taken is to be expected.


rhelmot:
Can you remove the large test files? If they're intended to actually be regression tests, can you put them in angr/binaries?

Lukas-Dresel:
@rhelmot of course. I haven't found a way to do automated testing of this given that it's pretty interlinked with the UI. I can just remove them, I'm okay with just having them in the branch if we need to too

Lukas-Dresel:
Fixed.

rhelmot:
tagging @etrickel since this mostly touches the multitrace stuff. other than that, lgtm

PR #150, opened 2020-03-24
DisasmView: Updating the view when labels are renamed.

ltfish:


ltfish:
It was implemented in the past, but the code got lost during the large refactor&merge.

PR #174, opened 2020-04-20
Support loading AFL bitmaps via angr URL scheme

ltfish:


rhelmot:
todo for when the crunch is over: move all this shit into the plugin. there shouldn't be any logic related to bitmaps, or even the _name_ for bitmaps, outside of the plugin.

PR #197, opened 2020-09-03
Support loading simple blobs

### LINKED ISSUE #127: Blob Loadng
Hello Angr-Team, 

I have a short question regarding the loading of blobs in angr-management. Is this possible or supported?

Best regards, 
Andy

mborgerson:
This patch simply adds the ability to load a file using the Blob backend in the event CLE refuses to load the file. GUI options are added to select CPU architecture, and specify appropriate base and entry point address. There are more options we can add to this in the future, but it does the job for now.

Fixes #127


ltfish:
YAY!

PR #206, opened 2021-01-25
Added initial fixes to locate an instruction from mouse location

mahaloz:
* related to the source-to-disass PR but opposite
* add tracking for disassembly view instruction when decompiling
* added a new instposmap for instruction locations making to position on angr-management display

### DIFF
--- a/angrmanagement/ui/views/disassembly_view.py
+++ b/angrmanagement/ui/views/disassembly_view.py
@@ -375,9 +375,10 @@ class DisassemblyView(BaseView):
     def decompile_current_function(self):
 
         if self._current_function.am_obj is not None:
-            curr_ins = None
-            if self.infodock.selected_insns is not None:
+            try:
                 curr_ins = next(iter(self.infodock.selected_insns))
+            except StopIteration:
+                curr_ins = None
 
             self.workspace.decompile_function(self._current_function.am_obj, curr_ins)
 
--- a/angrmanagement/ui/views/disassembly_view.py
+++ b/angrmanagement/ui/views/disassembly_view.py
@@ -380,7 +380,8 @@ class DisassemblyView(BaseView):
     def decompile_current_function(self):
 
         if self._current_function.am_obj is not None:
-            self.workspace.decompile_function(self._current_function.am_obj)
+            curr_ins = next(iter(self.infodock.selected_insns))
+            self.workspace.decompile_function(self._current_function.am_obj, curr_ins)
 
     def toggle_smart_highlighting(self, enabled):
         """
--- a/angrmanagement/ui/views/disassembly_view.py
+++ b/angrmanagement/ui/views/disassembly_view.py
@@ -167,7 +167,7 @@ class DisassemblyView(BaseView):
     # Events
     #
 
-    def keyPressEvent(self, event):
+    def keyReleaseEvent(self, event):
         key = event.key()
         if key == Qt.Key_G:
             # jump to window
@@ -193,12 +193,7 @@ class DisassemblyView(BaseView):
             # add comment
             self.popup_comment_dialog()
             return
-        super().keyPressEvent(event)
-
-    def keyReleaseEvent(self, event):
-        key = event.key()
-
-        if key == Qt.Key_Space:
+        elif key == Qt.Key_Space:
             # switch to linear view
             self.toggle_disasm_view()
             event.accept()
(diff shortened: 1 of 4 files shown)


mahaloz:
angr PR: https://github.com/angr/angr/pull/2496

Still fixing a bug to make the cursor update after a new position is given to it.

mahaloz:
Bug still present where cursor location is not actually updating for the new view after a new position has been set for the text cursor location.

angrmanagement/ui/workspace.py
```diff
@@ -264,7 +264,22 @@ def decompile_current_function(self, view=None):
...
+            doc = view.document
+
+            # get closest node for ins
+            new_text_pos = doc.find_closest_node_pos(curr_ins)
+
+            if new_text_pos is not None:
+                # set the new cursor position
+                textedit = view.textedit
+                cursor = textedit.textCursor()
+                cursor.setPosition(new_text_pos)
```
ltfish, line 279: You also need `textedit.setTextCursor(cursor)` and `textedit.setFocus()`.

ltfish (changes requested):
Two more things to change:

- Currently QCCodeEdit is set to read-only, which means at least on Windows its text cursor does not blink even if the widget has focus. We should change it to not read-only while disallowing any changes to its text.

- In DisassemblyView, move all logic in keyPressEvent() to keyReleaseEvent(). Otherwise they create weird conflicts (for example, pressing tab in disassembly view will jump to the pseudocode view, but then releasing tab will cause it to jump back).

angrmanagement/ui/views/disassembly_view.py
```diff
@@ -380,7 +380,8 @@ def display_function(self, function):
     def decompile_current_function(self):
 
         if self._current_function.am_obj is not None:
-            self.workspace.decompile_function(self._current_function.am_obj)
+            curr_ins = next(iter(self.infodock.selected_insns))
```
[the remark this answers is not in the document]
mahaloz: added check for None

angrmanagement/ui/workspace.py
mahaloz: done

angrmanagement/ui/views/disassembly_view.py
```diff
@@ -380,7 +375,11 @@ def display_function(self, function):
     def decompile_current_function(self):
 
         if self._current_function.am_obj is not None:
-            self.workspace.decompile_function(self._current_function.am_obj)
+            curr_ins = None
+            if self.infodock.selected_insns is not None:
+                curr_ins = next(iter(self.infodock.selected_insns))
```
ltfish, line 380: This is not the right fix. First of all, `self.infodock.selected_insns` is an `ObjectContainer`, so it will never be `None`. Second, the set wrapped inside `self.infodock.selected_insns` can be empty, which will cause `next(iter())` to fail with a `StopIteration`.

ltfish (changes requested):

angrmanagement/ui/workspace.py
```diff
@@ -264,7 +264,24 @@ def decompile_current_function(self, view=None):
...
+                textedit.setTextCursor(cursor)
+                textedit.setFocus()
+
+
+    def decompile_function(self, func, curr_ins, view=None):
```
ltfish, line 284: If you are changing the prototype of a method, you should change the docstring and type-annotate the arguments.

ltfish (changes requested):

angrmanagement/ui/views/disassembly_view.py
mahaloz: Got it, my bad. Fixed.

angrmanagement/ui/workspace.py
mahaloz: added typing and better docstring.

angrmanagement/ui/workspace.py
```diff
@@ -280,14 +280,16 @@ def position_cursor_on_decomp(self, view, curr_ins):
                 textedit.setTextCursor(cursor)
                 textedit.setFocus()
 
-
-    def decompile_function(self, func, curr_ins, view=None):
+    def decompile_function(self, func: Function, curr_ins: int, view=None):
         """
-        Decompile a function and switch to the pseudocode view.
-
-        :param Function func:   The function to decompile.
```
ltfish, line 271: You don't need to double-type-annotate `func`. Remove `Function`.

PR #210, opened 2021-01-29
Fix overly sensitive tabing on whitespaces

mahaloz:
* changed the search algorithm for tabing to look for the nearest node on both sides of cursor


mahaloz:
Pressing Tab Here:

demo

Now works like IDA does (goes to that line's statement start, not the line before it even if it's closer)

ltfish (approved):

PR #213, opened 2021-01-30
Added double clicking to memory location for constants

mahaloz:
* Double-clicking on strings works like IDA now:

https://user-images.githubusercontent.com/21327264/106343572-2868f080-624a-11eb-968b-b9233e46488b.mov


ltfish:
Good job!

PR #214, opened 2021-01-30
Aded highlighting of closure objects in decompiler view

mahaloz:
* added highlighting for objects like (), [], and {}


mahaloz:
Demo of IDA like highlighting:

https://user-images.githubusercontent.com/21327264/106345371-bf867600-6253-11eb-95e6-cadf8ed94f45.mov

PR #223, opened 2021-03-04
added better visualization for function_name nodes

mahaloz:
- added better highlighting for function name Nodes
    - preemptive upgrade before allowing function name renaming


mahaloz:
Correction: the renaming is concerning renaming variables across multiple-projects

ltfish (approved):

PR #225, opened 2021-03-04
Expose the console view

ConnorNelson:
We want access to the console view so we can extend it with our plugin.


rhelmot:
Why can't you use `view_manger.first_view_in_category('console')`? That's what the rest of the codebase does.

ConnorNelson:
Yes, of course 🤔
That was just a test. You passed.

PR #229, opened 2021-03-06
Add ability for plugins to instrument the pseudocode view

rhelmot:
Adds:

- ObjectContainer for currently selected node
- general instrumentation callback
- Context menu entries for currently selected node

### DIFF
--- a/angrmanagement/__main__.py
+++ b/angrmanagement/__main__.py
@@ -125,8 +125,9 @@ def start_management(filepath=None, use_daemon=False):
     splash.setWindowFlags(Qt.WindowStaysOnTopHint | Qt.FramelessWindowHint)
     splash.setEnabled(False)
     splash.show()
-    time.sleep(0.05)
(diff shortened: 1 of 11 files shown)


rhelmot:
And now, as a bonus for @mahaloz,

- Add context menu for functions, both in function table and pseudocode view
- Let Menu objects contain nested menus

mahaloz:
> And now, as a bonus for @mahaloz,
> 
>     * Add context menu for functions, both in function table and pseudocode view
> 
>     * Let Menu objects contain nested menus

@rhelmot, about this, any idea on how I could get a codeview generated for a function and get a reference to that view without actually changing the view of the code screen to that view? This is mostly just a hot fix for not having a decompiler kb since I need the kb associated to views for now.

rhelmot:
you probably want to loop through every currently loaded codeview and see if the function is cached/displayed.

Issue #231, opened 2021-03-11
Menu bar does not have some menus on MacOS


mahaloz:
On MacOS BigSur 11.2.1, with Intel Iris Plus Graphics, additions to the MenuBar sometimes do not appear in the native menu bar but will appear in the non-native menu.

ltfish:
Closing this issue since #232 has been merged.

PR #232, opened 2021-03-11
Added a temporary fix for Mac menu bars

mahaloz:
- Added a fix to not use native menu if on Mac
    - temporary usability fix until we can find the deeper problem in PyQT

Related to: #231

### DIFF
--- a/angrmanagement/ui/main_window.py
+++ b/angrmanagement/ui/main_window.py
@@ -1,5 +1,6 @@
 import os
 import logging
+import sys
 from typing import Optional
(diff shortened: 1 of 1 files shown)


mahaloz:
I don't think we actually need this for now, as long as we are adding new menu items to a plugin menu.

mahaloz:
Actually, this needs to still be implemented mac stuff still does not work even for plugin additions. Something strange is going on.

ltfish (approved):

PR #233, opened 2021-03-11
Added Binsync Plugin 

mahaloz:
- Added binsync plugin
- Moved all old binsync logic
- Made a new interface in view_manager to remove dead views with widgets

Note: uses MacOS fix from https://github.com/angr/angr-management/pull/232, which as of now is not merged yet. This will also need to be fixed after @ltfish redoes the decompiler interface -- but for now does no changes to decompilation caching.


mahaloz:
@rhelmot, feel free to add @ltfish to the review list if you think he should be here. Nothing big changed aside from the logic move and view manager.

ltfish (approved):

ltfish:
Fix linting!

rhelmot (approved):

rhelmot:
This looks great!

PR #243, opened 2021-05-14
Notify when the CFG is done generating

ConnorNelson:
No one seems to notify when the CFG is ready. It would be nice to know when it is ready. 

@rhelmot Is this how we should accomplish that?


rhelmot:
Yes, this is definitely the right way to do that. Thanks!

PR #258, opened 2021-05-25
Add more theming options

mborgerson:
* Changes some defaults on dark theme to be more sane (but really could use more tweaking)
* Distinguish constant operands with a color
* Adds an option for node rounding (perhaps superfluous, but I like it)
* Removes whitespace alignment in color scheme (it's nice but it causes churn)
* Fixes iPython color scheme update on theme change

Screenshot from 2021-05-25 16-08-12


ltfish:
Amazing!

PR #261, opened 2021-05-25
Allow C-c to kill angr-management

mborgerson:
For when you just want to kill it, allow C-c/SIGINT

### DIFF
--- a/angrmanagement/__main__.py
+++ b/angrmanagement/__main__.py
@@ -6,6 +6,7 @@ import threading
 import time
 import warnings
 import platform
+import signal
 
 def shut_up(*args, **kwargs):
     return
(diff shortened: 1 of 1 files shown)


ltfish:
You can use Ctrl+C to kill running python routines in the embedded IPython shell. Does this conflict with that?

I’m not sure we want to include such a hot key for a GUI program.

zardus:
The number of times I tried to ctrl-c angr-management to no avail... IMO,
we should quit when ctrl-c is hit. Individual processes should be killable
via the GUI (some sort of Jobs panel), ideally...

ltfish:
Does gedit or notepad quite when Ctrl-C is hit?

ltfish:
We can add a hot key that kills angr management as long as it does not conflict with existing hot keys or user habits.

mborgerson:
Sorry for the confusion, this is not to kill angr-management when Ctrl+C is pressed within the GUI, this is when you are at the console you launched angr-management and you press C-c to kill the running process like you would with other GUI apps launched from command line. e.g. if you start `gedit` then ctrl+c, gedit dies.

ltfish:
I would assume that it will not work on Windows? Can you wrap it inside a platform check?

mborgerson:
According to https://docs.python.org/3/library/signal.html this should work on Windows too, but I can wrap it if you prefer

ltfish:
Let's merge and see :)

PR #276, opened 2021-06-09
Support rendering multiline disassembly comments

### LINKED ISSUE #270: Multiline comments not rendered as multiple lines
Instead of showing as multiple lines, multiline comments are rendered with line feed escape sequence.

Screenshot from 2021-05-27 01-39-40

What I would expect instead would be empty space below the comment-bearing instruction, and the comment wrapping correctly at the next line. e.g.

```nasm
00000000   cmp     eax, ecx ; line 1
                            ; line 2
00000002   jnz     targ
```

rhelmot:
Closes #270


rhelmot:
image

PR #282, opened 2021-06-16
Support showing lifter IR, AIL in disassembly view

mborgerson:
Note: Depends on https://github.com/angr/angr/pull/2729

This patch adds the option to view different levels of abstraction of the target code in the disassembly views: machine code (what we already have), machine code with lifter IR, and AIL.

You can select which level you want in the combo box on the disassembly view status bar:
image

Switching over to 'Lifter IR' view shows the VEX statements inline with machine instructions for context:

image

And switching over to 'AIL' shows the AIL statements:

image

This works for both the Graph and Linear viewers:

image

image

The heart of this change is a new structured rendering widget called, for lack of a better name, `QBlockCode`. `QBlockCode` manages a tree of `QBlockCodeObj` objects that mirror whatever structure we are trying to display and renders it with Qt's `QTextDocument` engine. The design is extensible enough to apply to most of the structured code we would want to display and enables us to easily respond to individual events (like clicks on register identifiers, operators, etc.) and have fine-grain over the style of elements (e.g. highlighting, but we could do more like trace views).

Some things that can be improved, but are deferred to follow-up PRs:
* Better color schemes. I've mostly just used what we already have, except for addition of an IR color to more clearly differentiate it from machine code.
* Better input event handling. Right now clicking on an element sets a simple reference in `infodock`, like a cursor, mostly just for simple matching against similar elements, but it's not wired up to anything else except setting the current instruction selection.
* Context menus for different text elements.
* I had implemented a little animated coloring system that does a light highlight animation on identifiers when you click on them, but left it out of this PR for now. We can revisit it later.

While I was working on this I noticed some artifacts in the viewport after switching the scene around. These artifacts turned out to be caused by overpainting when drawing the block dropshadows. I've fixed that. I also see some artifacts from what looks like the block edges, but I haven't fixed that yet. It's very minor anyway, so honestly not a big priority.


ltfish:
LGTM - I'll wait for the tests to pass before hitting merge!

PR #286, opened 2021-06-18
Support opening multiple instances of a view

woadey:
Added a view menu option to create up to 26 new disassembly views.


angrmanagement/ui/workspace.py
```diff
@@ -3,6 +3,7 @@
 import traceback
 
 from PySide2.QtCore import Qt, QSettings
+from string import ascii_uppercase
```
mahaloz, line 6: @woadey I believe `string` is a built-in python library. Could you place this import next to `traceback` above, so that all imports of that kind are together.

angrmanagement/ui/workspace.py
```diff
@@ -143,6 +144,34 @@ def on_function_tagged(self):
...
+        :return:    None
+        """
+
+        if self.view_manager.disas_view_counter == 1:
+            for view in self.view_manager.docks:
+                if view.windowTitle() == 'Disassembly':
+                    view.setWindowTitle("Disassembly-A")
+                    self.view_manager.first_view_in_category('disassembly').caption = "Disassembly-A"
+                    break
+        if self.view_manager.disas_view_counter < 26:
```
mahaloz, line 162: Perhaps comments might be nice right here, as just from a glance it's unclear how you got `26` as view count number.

mahaloz (changes requested):
Looks good so far to me, but I think @ltfish might have some better feedback on your `view_manager` edits :).

rhelmot (commented):
This code gives me bad vibes but I can't describe how I would do it better. It just seems like it is not obvious from reading the code that it works - the enable_disas_button bool seems really sketchy and doesn't seem like it adds any major performance or safety improvements.

angrmanagement/ui/workspace.py
woadey: I'll make these changes once I am back in town on Monday!

woadey:
> This code gives me bad vibes but I can't describe how I would do it better. It just seems like it is not obvious from reading the code that it works - the enable_disas_button bool seems really sketchy and doesn't seem like it adds any major performance or safety improvements.

Wasn't too sure either, just added it to avoid unnecessary iterations (probably won't have any performance hindrances).

rhelmot:
I would appreciate it if we could get rid of both the disas_view_counter and the enable_disas_button variables. In the place of disas_view_counter, you should scan through the entire alphabet and create a view with the first name that is not used. I don't think enable_disas_button is even necessary - you can just do nothing or pop up an error if the user tries to create a 27th view. That's such a pathological case I don't think we need to cater to it in our design.

        self.disas_view_counter = 1
        self.enable_disas_button = True

woadey:
> I would appreciate it if we could get rid of both the disas_view_counter and the enable_disas_button variables. In the place of disas_view_counter, you should scan through the entire alphabet and create a view with the first name that is not used. I don't think enable_disas_button is even necessary - you can just do nothing or pop up an error if the user tries to create a 27th view. That's such a pathological case I don't think we need to cater to it in our design.

I am happy to get rid of the enable_disas_button variable, but I think it maybe difficult to remove the disas_view_counter. When the user closes out of a Disassembly view, the counter decrements which allows users to add a new view if they're under 26 disassembly views. If I remove the variable, then users could populate new disassembly views up to 26, but if they close out of the views they will not be able to add more.

mborgerson:
Why should a user be limited to 26 windows?

woadey:
> Why should a user be limited to 26 views?

I kept it to 26 for an easy A-Z naming convention. I also assumed having more views wouldn’t prove to add any functionality, but I can add it if there is. There also maybe more security implications if no limit is set.

mborgerson:
> > Why should a user be limited to 26 views?
> 
> I kept it to 26 for an easy A-Z naming convention. I also assumed having more views wouldn’t prove to add any functionality, but I can add it if there is.

26 views would be a lot, but I'm not sure I see any practical benefit to impose such a limit. Likewise, no other software I'm familiar with such as text editors, web browsers, etc. impose such a limit: you can create tabs until you run out of memory. If you want unique names with an A-Z naming you can do what spreadsheet apps do (e.g. A..Z, AA..AZ, BA..BZ, ...), or you could use natural numbers. Removing this limitation would simplify this patch a bit.

>  There also maybe more security implications if no limit is set.

What security implications in particular are you concerned about?

.gitignore
```diff
@@ -18,4 +18,4 @@ start.spec
 # enaml cache
 __enamlcache__/
 
-*.spec
+*.spec
```
ltfish, line 21: We need a new line here.

angrmanagement/ui/view_manager.py
```diff
@@ -194,6 +208,22 @@ def previous_tab(self):
         center_dockable_views = self.get_center_views()
         center_dockable_views[self.get_current_tab_id()-1].raise_()
 
+    def toggle_new_disas_button(self):
+        # Prevent double checking if already enabled
+        if self.enable_disas_button == False:
+            for menu_entry in self.main_window._view_menu.entries:
+                try:
+                    if menu_entry.caption == 'New Disassembly View':
```
ltfish, line 216: Let's not compare the menu captain. What if we have i18n support in the future?

angrmanagement/ui/workspace.py
```diff
@@ -143,6 +144,39 @@ def on_function_tagged(self):
...
+        central_widget and rename the views
+        accordingly
+
+        :return:    None
+        """
+
+        # Replace first disassembly view to appropriate naming convention 
+        if self.view_manager.disas_view_counter == 1:
+            for view in self.view_manager.docks:
+                if view.windowTitle() == 'Disassembly':
```
ltfish, line 159: Let's not rename existing tab titles. How about we always name the first Diasssembly View as "Disassembly-A?"

rhelmot:
I think the right thing to do to make both me and matt happy is to name the views with numbers instead of letters. That way there is never a need to prevent the user from creating a new view.

ltfish (dismissed):

ltfish:
We talked regarding eliminating the counter and using letters vs numbers. More changes will happen.

woadey:
@rhelmot @mborgerson @ltfish I have made the requested changes. Now, the default disassembly view's caption is 'Disassembly-1' and all new views follow this naming convention. (Per Fish's request, if the disassembly views range from 1-5, but 'Disassembly-3' gets deleted, a new 'Disassembly-3' will be populated before moving onto 6.)

There is some weird behavior going on with split_view() due to these additions which I have been looking into. I think the function itself needs changes - should I make those on this branch?

rhelmot:
Can you describe the weird behavior? To be honest, I'm not sure what the intended behavior of split_view is supposed to be.

woadey:
> Can you describe the weird behavior? To be honest, I'm not sure what the intended behavior of split_view is supposed to be.

I’m believe the intended behavior is to split the central_widget into 2 widgets (central_widget and central_widget2) thus allowing users to have split pane to look at multiple views simultaneously. 

The problem was the split_screen function uses the index gathered from get_current_id which looks at the central_widgets list, but split_screen uses the index to search the default_views list (which aren't the same lists). It works sometimes but not all. I rewrote the code on a local branch, but the weird behavior I am getting is that when I close out of the split screen view, it adds a new duplicated view for no reason

angrmanagement/ui/workspace.py
```diff
@@ -143,6 +144,29 @@ def on_function_tagged(self):
...
+    def new_disassembly_view(self):
+        """
+        Add a new disassembly view into the
+        central_widget with a unique id
+
+        :return:    None
+        """
+
+        dis_views = self.view_manager.views_by_category['disassembly']
+        dis_ids = [int(view.caption.split("-")[-1]) for view in dis_views]
```
ltfish, line 156: Instead of parsing the name of the view, can we add an `index` property to `BaseView`, then make `BaseView.caption` a property that always derives from `BaseView.base_caption` and `BaseView.index`?

ltfish (changes requested):

angrmanagement/ui/workspace.py
woadey: no problem, will add that now

angrmanagement/ui/views/disassembly_view.py
```diff
@@ -104,6 +103,20 @@ def setFocus(self):
     # Properties
     #
 
+    @property
+    def caption(self):
+        return f'{self.base_caption}-{self.index}'
+
+    @caption.setter
+    def caption(self, cap):
```
ltfish, line 111: Let's get rid of `caption.setter`. `caption` should be a read-only property that is always constructed from `self.base_caption` and `self.index`. You'll want to add `self.base_caption = 'Disassembly` in `DisassemblyView.__init__` and `self.index = 1` in `BaseView.__init__`.

Also, Please move `caption` from `DisassemblyView` to `BaseView` so all views are treated equally.

ltfish (dismissed):

angrmanagement/ui/views/disassembly_view.py
woadey: updated!

angrmanagement/ui/view_manager.py
```diff
@@ -136,12 +135,20 @@ def get_center_views(self):
     def first_view_in_category(self, category) -> Optional['BaseView']:
         """
         Return the first view in a specific category.
+        If the category is disassembly, return the current
+        category view else return the first view in
+        disassembly.
```
ltfish, line 140: Why do we want to do this? The method name is `first_view_in_category`, which means we should _always_ return the first view in each category, not the current view..

ltfish (changes requested):
One last fix required before I can hit merge!

angrmanagement/ui/view_manager.py
woadey: The reason I implemented this in the first_view_in_category function was because its used quite a bit, especially in generating the content on the disassembly view. It was a simple workaround to a larger issue I think. I've been working on trying to have a list constantly updated according to the current view history, but haven't been able to figure out how to detect when a user changes tabs (hence updating the view history list). 

I also tried to just copy the logic of this function and create a separate one and replace all instances of `self.first_view_in_category('disassembly)` but I haven't been able to determine what is going wrong.

Still working on this change.

ltfish: > The reason I implemented this in the first_view_in_category function was because its used quite a bit, especially in generating the content on the disassembly view. It was a simple workaround to a larger issue I think. 

What is the issue that you are trying to address?

woadey: > > The reason I implemented this in the first_view_in_category function was because its used quite a bit, especially in generating the content on the disassembly view. It was a simple workaround to a larger issue I think.
> 
> What is the issue that you are trying to address?

There was a lot of code using the `first_view_in_category` function, but if we are starting to add multiple views per view type, this function will become obsolete for anything other than the initialization of the default views.

rhelmot: I think the correct behavior of `first_view_in_category` is to return the _most recently accessed_ view of that category. That would fall in line with the way the function is used and the way similar functionality works in IDA. I do however agree that there are a lot of spurious uses of the function in general, in places where it would be a lot cleaner for an object to have an explicit reference to it parent view.

ltfish: > I do however agree that there are a lot of spurious uses of the function in general, in places where it would be a lot cleaner for an object to have an explicit reference to it parent view.

Let's not touch this function in this PR. We can make a separate PR for changing the behavior of this function.

Moreover, you should never change the docstring and the code of a function so that its behavior conflicts with what its name suggests.

woadey: > > I do however agree that there are a lot of spurious uses of the function in general, in places where it would be a lot cleaner for an object to have an explicit reference to it parent view.
> 
> Let's not touch this function in this PR. We can make a separate PR for changing the behavior of this function.
> 
> Moreover, you should never change the docstring and the code of a function so that its behavior conflicts with what its name suggests.

I went ahead and created a separate `current_view_in_category` function, but I changed code around the calling function of `first_view_in_category`. Now, anytime there is _only_ the first view of the Disassembly view, it will call `first_view_in_category` as it did before. But as soon as a new view is added, the `current_view_in_category` function will be called instead.

ltfish:
LGTM. Resolve the conflict and then I'll hit merge! Good job!

ltfish (approved):

ltfish (approved):

PR #287, opened 2021-06-18
add `step_callback` for plugin

yuzeming:


rhelmot (changes requested):
First, some style guidelines:
the `"""` for a docstring should always be on its own line, like this:

```
"""
Collect a list of functions from the active plugins corresponding to the base_func from BasePlugin.
"""
```

When doing type annotations, there should be a space between the colon and the type.

Finally, I am not convinced any of your changes to plugin_manager are necessary. Is there really such a performance overhead associated with crawling the plugin classes for the appropriate functions that it can't be done during execution? Is there something wrong with the way we're doing error handling?

rhelmot (approved):

PR #288, opened 2021-06-18
Fix: stop binsync from loading without install

mahaloz:
- stops loading of binsync widows and callbacks if its not installed


mahaloz:
@ltfish I'm not so sure that my fix is such a great idea. Maybe I should consult @rhelmot since she has a better vision of what kind of permissions plugins should have (including how to remove themselves from view).

rhelmot:
Probably the best way to do this is to in the binsync plugin module itself, if binsync is not installed then `del` the plugin class :)

mahaloz:
> Probably the best way to do this is to in the binsync plugin module itself, if binsync is not installed then `del` the plugin class :)

Do this before or after `super().__init__()`? @rhelmot

rhelmot:
```
try:
    import binsync
except ImportError:
    binsync = None

class BinsyncPlugin(BasePlugin):
  ...

if binsync is None:
    del BinsyncPlugin
```

mahaloz:
> ```
> try:
>     import binsync
> except ImportError:
>     binsync = None
> 
> class BinsyncPlugin(BasePlugin):
>   ...
> 
> if binsync is None:
>     del BinsyncPlugin
> ```

Seems to be illegal:
```
Traceback (most recent call last):
  File "/Users/mahaloz/github/angr-dev/angr-management/angrmanagement/plugins/plugin_manager.py", line 92, in activate_plugin
    plugin = plugin_cls(self.workspace)
  File "/Users/mahaloz/github/angr-dev/angr-management/angrmanagement/plugins/binsync/binsync_plugin.py", line 22, in __init__
    del BinsyncPlugin
UnboundLocalError: local variable 'BinsyncPlugin' referenced before assignment
```

mahaloz:
nvm

mahaloz:
Honestly @rhelmot, I don't know how this works. Code inside a class, outside of an `__init__` runs when? When the class is first imported?

rhelmot:
You need to unindent your del code one more time. It should run on module import.

mahaloz:
@rhelmot this makes a lot more sense. That is slick. amazing.

rhelmot (approved):

PR #300, opened 2021-06-18
Feat: allow multiple functions passed in a context menu

mahaloz:
- added passing of multiple function selections for the function table context menu

### DIFF
--- a/angrmanagement/ui/widgets/qccode_edit.py
+++ b/angrmanagement/ui/widgets/qccode_edit.py
@@ -111,7 +111,7 @@ class QCCodeEdit(api.CodeEdit):
             # decompiled function name in selection
             self._selected_node = under_cursor
             mnu.addActions(self.function_name_actions)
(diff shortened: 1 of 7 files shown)


angrmanagement/ui/widgets/qccode_edit.py
```diff
@@ -111,7 +111,7 @@ def get_context_menu(self):
             # decompiled function name in selection
             self._selected_node = under_cursor
             mnu.addActions(self.function_name_actions)
-            for entry in self.workspace.plugins.build_context_menu_function(self.workspace.instance.kb.functions[under_cursor.name]):
+            for entry in self.workspace.plugins.build_context_menu_functions(self.workspace.instance.kb.functions[under_cursor.name]):
```
rhelmot, line 114: Should this be `([self.workspace...` instead of `(self.workspace...`? i.e. pass a list?

mahaloz: @rhelmot yup, good catch

PR #322, opened 2021-06-21
Feat: BinSync UI

mahaloz:
- Added the GUI of BinSync as well as reorganized all BinSync code. 

Note: new GUI does not have full support for the new BinSync pushing and pulling, and thus will be added with another PR for both angr-management and angr (for the new BinSync API).

### DIFF
--- a/angrmanagement/data/sync_ctrl.py
+++ b/angrmanagement/data/sync_ctrl.py
@@ -36,6 +36,7 @@ class SyncControl:
     """
     def __init__(self, instance):
         self.instance = instance
+        self.client = None
 
         self.project = None
         self.users_container = ObjectContainer([], notes="All users in the current team.")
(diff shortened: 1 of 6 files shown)


mahaloz:
@ltfish I need help fixing this error in the CI:
```
LINT FAILURE: new file angrmanagement/plugins/binsync/info_tables/func_info_table.py lints at 9.32/10.00. Errors:
... angrmanagement/plugins/binsync/info_tables/func_info_table.py:5:0: E0401: Unable to import 'binsync.data' (import-error)
```

Do I just need to update the `setup.py` for `binsync`?

rhelmot:
I don't believe binsync is installed during our CI runs...

I think you need to use the optional import idiom for this.

mahaloz:
> I don't believe binsync is installed during our CI runs...
> 
> I think you need to use the optional import idiom for this.

I decided to go the `try: import XXX` route.

PR #325, opened 2021-06-23
Properly support high-DPI displays.

ltfish:
We should never draw text manually using `painter.drawText()` since it's impossible to calculate the bounding rect of the text before possessing a painter instance. Use `QGraphicsSimpleText` instead. This is the root cause for the ugly disassembly graph on MacBooks (since the text lengths that we calculate in `Conf` are different from the actual text lengths in the `painter` when painting the graph).

Reference: https://stackoverflow.com/questions/34828777/why-does-qpainter-seem-to-scale-fonts-unpredictably-when-using-drawtext-on-a-qim

TODO:

- [x] Migrate `QUnknownBlock`.
- [x] Migrate `QDepGraphBlock`.
- [x] Migrate `QProximityGraphBlock`.
- [x] Drop `devicePixelRatioF` and `QCachedGraphicsItem._cached_device_pixel_ratio`. Maybe also `QCachedGraphicsItem._container`.
- [x] Drop `QCachedGraphicsItem.pixelToPoint`.

### DIFF
--- a/angrmanagement/__main__.py
+++ b/angrmanagement/__main__.py
@@ -120,10 +120,19 @@ def start_management(filepath=None, use_daemon=False):
 
     from PySide2.QtWidgets import QApplication, QSplashScreen, QMessageBox
     from PySide2.QtGui import QFontDatabase, QPixmap, QIcon
-    from PySide2.QtCore import Qt
+    from PySide2.QtCore import Qt, QCoreApplication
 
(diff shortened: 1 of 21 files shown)


ltfish:
An example of the ugly disassembly graph on MacBooks: The screenshot in https://github.com/angr/angr-management/pull/323

Here is the expected view:

image

PR #329, opened 2021-06-24
Add Plugin to log renamed node stats

Adersh97:


angrmanagement/ui/main_window.py
```diff
@@ -565,6 +565,7 @@ def _save_database(self, file_path):
         angrdb.dump(file_path)
 
         self.workspace.instance.database_path = file_path
+        self.workspace.plugins.handle_project_save(file_path)
```
rhelmot, line 568: This callback should be put before the database is serialized, in case users want to make any changes to the data to be saved.

angrmanagement/plugins/log_reverse_engineering/log_reverse_engineering_plugin.py
```diff
@@ -0,0 +1,133 @@
...
+    from slacrs import Slacrs
+    from slacrs.model import VariableRename, FunctionRename, ReverseEngineeringProgress
+except ImportError as ex:
+    Slacrs = None  # type: Optional[type]
+    VariableRename = None  # type: Optional[type]
+    FunctionRename = None  # type: Optional[type]
+    ReverseEngineeringProgress = None  # type: Optional[type]
+
+
+class LogReverseEngineeringPlugin(BasePlugin):
```
rhelmot, line 13: You need to add a few lines to this file to prevent the plugin from being loaded if slacrs is not installed. see https://github.com/angr/angr-management/pull/288 for instructions.

rhelmot (changes requested):

rhelmot (approved):

rhelmot:
Actually, can you do the same thing as you did to the logfatigue plugin and add this to the plugin blacklist too?

twizmwazin (approved):

Adersh97:
Yes i have added.

rhelmot:
As per the discussion in the other issue, you should actually remove the `del` logic and instead add a thing in `__init__` which will raise a useful error message if slacrs is not installed.

Adersh97:
> As per the discussion in the other issue, you should actually remove the `del` logic and instead add a thing in `__init__` which will raise a useful error message if slacrs is not installed.

Changes Made.

PR #331, opened 2021-06-24
stop LogFatigue plugin from initializing if required library is not installed

Adersh97:


rhelmot (approved):

ltfish:
Hang on - I'm not sure why this is necessary since we are controlling which plugins are enabled using the configuration file.

rhelmot:
imo the right way to do it is to control the plugin loading with _both_ the configuration file and preventing the plugin from loading if its dependencies aren't installed. I'll ask for the other plugin to be added to the blacklist too.

ltfish:
> and preventing the plugin from loading if its dependencies aren't installed.

It should actually complain. Imagine that the user wants to enable this plugin but some dependencies are missing, what should happen is that we try to load it, and then complain (e.g., raise an exception that can be caught outside and a messagebox can be popped up) to the user that the plugin failed to load and why.

rhelmot:
So.... the right thing to do is to _just_ use the try: import idiom and to explicitly say `if dependency is None: raise Exception("Please install dependency")` in the plugin initializer. That way the error is deferred until instantiation and is explicit about what it needs. I think we already show plugin instantiation errors to the user.

PR #332, opened 2021-06-27
Add keybindings to zoom the pseudocode with Ctrl-minus and equals.

zardus:
Zooming was already possible with Ctrl-Scroll, but this additional behavior matches the disassembly view.


ltfish:
Merged. Thanks!

Issue #339, opened 2021-06-29
Navigating between functions in disassembly view resets the zoom


zardus:
This makes me sad.

PR #340, opened 2021-06-29
Partial fix for #339: preserve zoom when jumping in disas view.

zardus:
Unfortunately, the screen is positioned without taking the zoom into account...


zardus:
@rhelmot , it looks like you implemented the zoom stuff. What do you think of this?

rhelmot (approved):

PR #342, opened 2021-06-29
Check for non-printable characters before printing a string to UI

WGibbs15:


ltfish:
Lint the code and then I'll hit merge!

PR #356, opened 2021-06-30
remove SrcAddrAnnotation

yuzeming:
I remove the second column of QConstraintViewer which show the SrcAddr of a constrain.
It caused crash when users hit explorer button double times.

This function should be added in angr, not there.


rhelmot:
You can get the source address of a constraint by finding the appropriate SimAction for it in state.history.actions!

PR #368, opened 2021-07-02
implementation of MultiTrace + Heatmap w/ SLACRS support

capysix:


rhelmot:
I don't see any problems with this. Fix lint and I'll hit merge.

twizmwazin:
Should the print statements be swapped out for log messages?

rhelmot:
Oops, didn't see that. Yes, they should be. Use `workspace.log()`.

capysix:
Would it be better for me to rebase and squash all my linting fixes before merging?

rhelmot:
We have a handy squash-merge button for this exact purpose :)

Issue #348, opened 2021-06-29
Instruction annotations hold references to views


rhelmot:
@yuzeming I can't assign you but this issue is for you.

QInstructionAnnotation has a self.disasm_view and a self.symexec_view. Ideally it should have neither of these.

The symexec view reference definitely needs to go - it is currently possible to trigger a crash by creating an annotation that refers to a symexec view and then closing that symexec view. Consider using _get_or_create_symexec_view() instead.

For the disasm view, it's not strictly a problem yet but it could be in the future if we even wanted to render blocks outside of a disassembly view. At the very least, it should make sure it has a reference to its parent disasm view instead of just using first_view_in_category, since there can be multiple disassembly views.

PR #369, opened 2021-07-02
Remove `symexec_view` ref in `QInstructionAnnotation`

yuzeming:
Fix https://github.com/angr/angr-management/issues/348

### DIFF
--- a/angrmanagement/ui/widgets/qinst_annotation.py
+++ b/angrmanagement/ui/widgets/qinst_annotation.py
@@ -18,15 +18,14 @@ class QInstructionAnnotation(QGraphicsSimpleTextItem):
     background_color = None
     foreground_color = None
     addr = None
-    container = None # type: QBlockAnnotations
 
     @property
     def disasm_view(self) -> 'DisassemblyView':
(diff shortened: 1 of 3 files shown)


angrmanagement/ui/widgets/qinst_annotation.py
```diff
@@ -21,6 +21,14 @@ class QInstructionAnnotation(QGraphicsSimpleTextItem):
     foreground_color = None
     addr = None
     _config = Conf
+        
+    @staticmethod
+    def get_disasm_view() -> 'DisassemblyView':
+        return GlobalInfo.main_window.workspace.view_manager.first_view_in_category("disassembly")
```
rhelmot, line 27: This isn't right - if there are two disasm views open, the annotations in one of them can return the other one. You need to explicitly pass in the parent disassembly view as a constructor parameter.

yuzeming: I move `disasm_view` ref to the container `QBlockAnnotations`.

rhelmot (approved):

PR #382, opened 2021-07-07
Show variables for AIL Register/Load

mborgerson:
This patch:
* Fixes display of variables in AIL view, controllable by disassembly view options
* Adds very basic renaming support for variables, with scaffolding to support interactivity on other items
* Show AIL call return expression and arguments

image


ltfish (approved):

ltfish:
Exciting!

PR #386, opened 2021-07-10
Sync with the console when select state/stash in SimgrViewer

yuzeming:
Add click func for state items in `SimgrViewer`

图片

based on @filipkilibarda 's work.


ltfish:
Let's avoid pushing `simgr`, `s`, and `states` to the outer namespace since they are commonly used by users. Pushing them to the outer namespace will overwrite existing declarations of these variables.

My suggestion: Assign them to a global class, e.g., `S`. So they can be accessed via `S.simgr` or `S.states`.

yuzeming:
Replaced by https://github.com/angr/angr-management/pull/419
I closed this one.

PR #398, opened 2021-07-15
Add history navigation buttons to disassembly/code views

### LINKED ISSUE #291: Back and Forward buttons for Disassembly Views

mborgerson:
Fixes #291 

Works basically like a web browser, with nav buttons in the status/control bar of the view. Adds dropdowns to each nav buttons to allow navigating the history explicitly.

The alternative that I had considered was moving the nav buttons to the main window toolbar, but that would add complexity with focus tracking--especially when windows are undocked. In the future we could explore moving interaction buttons up to the main window when docked and only show them on the undocked version otherwise, but I think this solution works well enough for now.

Also in the future I'd like to improve this to use more human-friendly names, for now it just uses explicit addresses.

image


ltfish:
Awesome!

PR #401, opened 2021-07-16
Feat: Testcases for BinSync

mahaloz:


mahaloz:
Added testing for binsyncs:
- function renaming hooking & syncing
- stack variables in decompilation hook & syncing

In the latter of the two tests, I do a significant amount of just normal stack variable renaming, which should probably also be it's own testcase. Lmk if you want it in a different PR @rhelmot or @ltfish.

mahaloz:
blocked on merging this until I get `binsync` in the deps for CI.

ltfish (approved):

Issue #402, opened 2021-07-19
Disable deep linking by default


twizmwazin:
By default, we should not prompt the user to configure the link handler as this feature is not intended to be used by most users.

ltfish:
Addressed in #412.

PR #412, opened 2021-07-22
Only prompt registering angr:// URL scheme in CHECRS mode.

ltfish:


twizmwazin:
Does this address #402?

PR #415, opened 2021-07-24
Fix minimap performance for large functions

mborgerson:
Functionally the minimap works exactly as before, just faster. What's changed:
* Restructures minimap to use QGraphicsView/QGraphicsItem
* Caches the displayed scene in an image

Also:
* Remove dead `QGraphObject`


ltfish:
Amazing! It is indeed much faster.

PR #419, opened 2021-08-01
add `QFileDescriptorViewer`

yuzeming:
图片

to dump fd for found states.


yuzeming:
and should we remove the `Temps` tab?

ltfish:
> and should we remove the Temps tab?

Any reasons?

PR #427, opened 2021-08-08
Fix 'NoneType' object has no attribute and 'utf-8' codec can't decode byte 0x80

yuzeming:


yuzeming:
resubmit PR, closed this one.

PR #478, opened 2021-09-14
Add edit feature to Hex view

### LINKED ISSUE #463: Support editing within hex view
Should be able to create a patch from within the hex view.

In Ida, this works like:
* Right click on a byte, select Edit (F2)
* Make your changes
* Right click, select 'Apply changes' (F2)
* Change appears in 'Patched bytes' subview

mborgerson:
Add edit feature to Hex view. Simply place the cursor where the edit should happen and begin typing. Edits are stored as patches in the knowledge base and can be split, merged, and removed via context menus. Patches are naturally synchronized with Patches view. Basic copy-paste support is added.

There is definitely room for improvement here, but I've deferred it for now. Some thoughts:
* Upon edit, running targeted analyses to, for instance, let edits to code be reflected in disassembly view. This will require more thought.
* Integrating into system clipboard to support copy of bytes (optionally formatted as C/Python code, Hex, binary, etc) out or in; similarly permit import/export of file at location.
 
Demo:

https://user-images.githubusercontent.com/8210/133179617-e5029a20-3e64-450d-bb70-fc8d2c6cc18d.mp4

Closes #463

### DIFF
--- a/angrmanagement/ui/dialogs/input_prompt.py
+++ b/angrmanagement/ui/dialogs/input_prompt.py
@@ -1,0 +1,54 @@
+from PySide2.QtWidgets import QDialog, QVBoxLayout, QHBoxLayout, QLabel, QLineEdit, QDialogButtonBox
+
+
+class InputPromptDialog(QDialog):
+    """
+    A generic dialog to prompt for text input.
+    """
+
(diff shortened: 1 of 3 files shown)


ltfish:
Amazing!

PR #547, opened 2021-12-23
Add generic Debugger interface, simulation debugger

mborgerson:
* Adds scaffolding for a generic Debugger interface
* Adds a new stack and registers view
* Adds a new 'Debug' toolbar, that extends/replaces the existing Simgr toolbar allowing the user to use the buttons to to control the selected debugger.
* The debug toolbar contains a combo box letting you select between different debuggers, and the state of each debugger is reflected in Disassembly, Stack, and Registers view.
* The simulation debugger operates on the selected Simulation Manager in Symbolic Execution tab. You can switch between them, select a state, etc. and it will be reflected in associated views.

https://user-images.githubusercontent.com/8210/147299224-c85d0e26-2d76-4bc7-87c7-e1fdc3459f78.mp4

This has a long way to go to become more useful, but it's a start. What's not here, but are planned for follow-up PRs:
* GDB debugger, using avatar2/symbion integration, + gdb console integration
* Trace debugging (needs more refactoring to make it play nice with existing trace plugin)
* Trace exploration, visualization
* Integrated trace+GDB for timeless debugging
* Thread views
* Breakpoints/watchpoints
* Better ways to view/mutate memory (e.g. letting hex viewer view state's memory, not just memory from loader)
* Making registers/stack editable
* Stack view improvements
  * Arbitrary stack depth
  * Address decoding
  * Argument decoding based on debug info
  * Traversal of stack frames


ltfish:
Tests were failing due to incompatibility between the latest `qtpy` and `pyqodeng.core`. I fixed the issue, and the CI should be passing.

PR #568, opened 2022-01-21
add `enabled_tabs` in Conf

yuzeming:
So we can remove some tabs when we don't need them in the experiment.


ltfish:
Can you fix linting?

ltfish:
Merged. Thanks!

PR #591, opened 2022-03-22
[WIP] AIL Graph Refactor

mahaloz:
- AIL graph will now be based on decompilation opts instead of a new clinic graph. This also means we will now have variable unification in the AIL graph. 

Not ready for merging


mahaloz:
@ltfish any eta on when we might have a different Job system in AM? This PR is still waiting on the ability to start a decompilation job while the display job is running for the AIL graph.

PR #600, opened 2022-03-30
[WIP] PySide6 Support

mahaloz:
## Background
The current and hardest problems for this port has to do with a fork of pyqode that we use:
https://github.com/ltfish/pyqodeng.core

This repo depends on `qtpy`, which we "use" for Qt version agnostic imports. QtPy currently does not support many imports that we still use in PySide6. An example is the `QUndoCommand` which was moved from Qt.Widgets to Qt.Gui in PySide6. However, `qtpy`, does not fix this case: 
https://github.com/spyder-ide/qtpy/blob/6199be5ad3d4ed16d69ee1ac48f45e59cdc8436b/qtpy/QtWidgets.py#L39

There are **many** such cases of this that I have tracked. The real problem is that we are even still trying to use an agnostic platform like this. 

## Proposal 
To address these issues I propose we do three things:
1. Remove use of `qtpy` from the entirety of angr-management
2. Support only PySide6 in angr-management 
3. Do a full refactor of our fork of pyqodeng to no longer use `qtpy` and only use PySide6

Currently, I have started by fixing some imports that have been deprecated or moved for PySide6, but now this PR waits on changes to pyqode fork for PySide6 use only.

### DIFF
--- a/angrmanagement/__init__.py
+++ b/angrmanagement/__init__.py
@@ -1,11 +1,11 @@
 __version__ = "9.2.0.dev0"
 
 # Hack used to work around the slow-responsiveness issue with the GUI
-# PySide2 5.14.2 solves this problem but it introduces other bugs
+# PySide6 5.14.2 solves this problem but it introduces other bugs
 # See https://bugreports.qt.io/browse/PYSIDE-803
 try:
-    import PySide2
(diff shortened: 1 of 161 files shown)


twizmwazin:
Superseding this with #733

PR #629, opened 2022-04-12
Plugin to display coverage from traces

adamdoupe:
Not ready yet, but putting this here so @ltfish can see it.


ltfish:
@adamdoupe Can you please properly implement the caching mechanism and avoid having `state={}` in the function argument list?

PR #634, opened 2022-04-13
Refactor Decompilation Callbacks and add further support

mahaloz:


angrmanagement/plugins/base_plugin.py
```diff
@@ -172,63 +172,43 @@ def step_callback(self, simgr:SimulationManager):
...
     #
 
-    def decompile_callback(self, func):
-        """
-        A callback that is called *right after* the decompiler is run on a function. You can access the current codegen
-        with ``self.workspace.instance.kb.structured_code[(func.addr, 'pseudocode')]``
-        :param func:        angr Function that was just decompiled
-        """
+    def handle_stack_var_renamed(self, func, offset, old_name, new_name):
+        return False
```
ltfish, line 176: There are more than just `stack_var` and `global_var` in decompilation output. For example, do you consider `reg_var` at all?

Ideally we should have a callback for each of the following: `stack_var`, `global_var`, `reg_var`, and `other_var`.

ltfish (changes requested):

angrmanagement/plugins/base_plugin.py
mahaloz: @ltfish I don't really consider it because I don't see a clean way of making a callback. For instance, a "reg_var" is a SimVar, which could be anything. I don't see a simple API that allows someone to know when a register variable changed without them knowing how to parse the arcane angr SimVar stuff. In addition, a register var and an "other" var don't really exist in code, making it even harder to parse here.

ltfish: I must be missing something here. Why do you think a "reg_var" is a SimVar, instead of a `SimRegisterVariable`?

mahaloz:
@ltfish I've added a `handle_other_var_renamed` and retyped function. I think this should suffice.

ltfish (approved):

PR #635, opened 2022-04-14
Update BinSync to version 2.2.0

mahaloz:


mahaloz:
I would love to change how we do this, so this will be the last time I update code manually like this. Ideally we should have some type of remote loading plugin system @rhelmot.

ltfish:
Redesigning the plugin loading is definitely on our plate but it's not of a high priority.

mahaloz:
@ltfish if you want a quick and dirty hack, we could always just submodule binsync

ltfish:
Please no. I'm about to remove the submodules in this repo...

PR #684, opened 2022-05-19
Generalize socket_config.py 

### LINKED ISSUE #683: MacOS broken due to non-present sockets in socket_config.py
This change was added to support windows:
https://github.com/angr/angr-management/commit/c3d963ee4677b50d50172c87d2dd70be1575814c

However, some of those sockets, such as `AF_CAN`, are not present on MacOS, while others, such as `AF_UNIX`, are on MacOS, which causes anger-management to crash on MacOS.

Rather than switching based on the platform, let's decided based on what actually in the `socket` module.

adamdoupe:
Fix https://github.com/angr/angr-management/issues/683

Generalize the socket_config.py to only load the socket families that we want that are actually present in the `socket` module, so we need to do no platform testing.

Fixes #683.

### DIFF
--- a/angrmanagement/ui/dialogs/socket_config.py
+++ b/angrmanagement/ui/dialogs/socket_config.py
@@ -1,6 +1,5 @@
 import socket
 import base64
-import platform
 
 import claripy
 from PySide2.QtGui import QIntValidator, QContextMenuEvent, QColor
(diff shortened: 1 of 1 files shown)


ltfish:
Thanks!

PR #699, opened 2022-06-14
Skip highlighting CFB item of unknown size in Hex view  

tr1shul:
Applying the patch skips over MemoryData items with unknown size.


mborgerson:
Just a nit about commit message. Prefer brief commit message title about what the patch does. Like:

`Skip highlighting CFB item of unknown size in Hex view`

Then in the message body we can add more details about the issue the patch is correcting, if necessary

ltfish (approved):

PR #731, opened 2022-09-01
fix dict(None)

yuzeming:


rhelmot:
I feel like the right way to do this is to initialize env_config to an appropriate default value (empty list?) instead of checking for a pathological value later.

yuzeming:
updated

Issue #729, opened 2022-08-07
Add re-compilation feature to decompilation view


mborgerson:
To achieve decompilation of a binary closer, or possibly identical, to original source and to discover subtle bugs in re-implementation, I've automated a simple workflow of adjusting source + compiling + binary diffing. I think this could make for a useful angr-management feature.

A new option to the pseudocode view would be added that allows the user to enable re-compilation and specify compiler options. Whenever a tweak is made that prompts pseudocode regeneration, it is automatically recompiled and and a diff against the original is displayed.

What my script displays: ![Screenshot from 2022-08-06 16-43-13](https://user-images.githubusercontent.com/8210/183269824-69d4d601-903a-4b93-ad73-84be84ea7942.png)

PR #739, opened 2022-10-12
Refactor: Instance Based Views

mahaloz:
In most cases, a View (like disassembly view), is created using the `workspace` with the use of the `instance` object. In effect, the instance object is a class that contains all the data of an angr project. To allow angr-management to have more than a single version of a View open, we must make views based on `instance` not on workspace. 

In this update, I refactor A LOT of code so that most references to workspace are exchanged with instance. This allows us, for instance, to load two binaries in at the same time in angr-management and do cool comparison stuff.

Using this PR code, here is a preliminary implementation for #729, to allow for recompilation and object difference viewing. Attached is a screen shot of this. I may move the plugin code that does this to a different PR, but for now it shows a nice way of why decoupling views from Workspace is essential and how you can test it. 

## Binary Diffing on ASM Level
recompilation


mahaloz:
@mborgerson, you may be interested in how this is done :)

zardus:
Wait, what... You made a-m support multiple angr projects??? It seems awesome in retrospect, but that's one of these ideas that simply never occurs to people!

mahaloz:
@zardus No Binary is an Island. Hehe. But as you may have realized, being able to support multiple binaries loaded into angr management at the same time makes for some interesting implementation opportunities.

One I discussed with @ltfish was the ability to load library binaries in with a main loader and correct xrefs to them for some sick reversing of multi-bin targets.

mahaloz:
For any who review this PR, please don't merge without talking to me first. I will likely move the BinDiff code to another PR.

ltfish:
> Wait, what... You made a-m support multiple angr projects??? It seems awesome in retrospect, but that's one of these ideas that simply never occurs to people!

It has been the idea since the very early days of angr management, and the workspace/instance structure has always been designed to make that feasible. It's just that I never had enough energy to implement the support for it. Now that we have @mahaloz ... ;)

mahaloz:
@ltfish the plugin code is moved to #740

ltfish (approved):

PR #754, opened 2022-10-25
Add plugin description files. Delay loading plugins.

ltfish:


ltfish:
Linting fails because we are not ignoring PySide6. [This PR](https://github.com/angr/angr-dev/pull/1660) fixes it.

twizmwazin (commented):
Previously we had discussed a plugin system that would take advantage of python packaging standards. Do we still want to do that? If so, it would make many of the plugin.toml fields redundant. If we choose to go this route and keep plugins separate from python packages, do we want to have a field to declare plugin dependencies? We could pretty easily just wrap pip to install them, assuming a pip environment.

ltfish:
> Previously we had discussed a plugin system that would take advantage of python packaging standards. Do we still want to do that?

Maybe, but it will take a lot more time to think about and implement.

> If we choose to go this route and keep plugins separate from python packages, do we want to have a field to declare plugin dependencies?

We do. It can be added as another field to `[plugin]`, but it does not make sense before we settle on how the distribution (and installation) of plugins will work.

ltfish:
@twizmwazin FYI https://github.com/angr/angr-management/issues/757

PR #766, opened 2022-10-31
Drop PySide6 constraint

mborgerson:
Before upgrading we need:
* [qtpy](https://github.com/spyder-ide/qtpy) to make a release
* ~~[qtconsole](https://github.com/jupyter/qtconsole) to merge [my PR](https://github.com/jupyter/qtconsole/pull/561) and make a release~~

In the mean time you can:
```
pip install -U PySide6
pip install https://github.com/spyder-ide/qtpy/archive/refs/heads/master.zip
```

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -17,7 +17,7 @@ install_requires =
     angr[angrDB] == 9.2.26.dev0
     qtconsole
     ipython
-    PySide6 < 6.4.0
+    PySide6 >= 6.4.0.1
     tomlkit
     pyxdg
     requests[socks]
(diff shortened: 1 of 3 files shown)


setup.cfg
```diff
@@ -17,7 +17,7 @@ install_requires =
     angr[angrDB] == 9.2.26.dev0
     qtconsole
     ipython
-    PySide6 < 6.4.0
+    PySide6 >= 6.4.0.1
```
twizmwazin, line 20: Maybe we want this to be `~= 6.4.0` instead of assuming 6.5 will just work?

mborgerson: I don't expect even 6.4.1 to work tbh

twizmwazin: `~= 6.4.0.1` would be 6.4.0.x but not `6.4.1`

PR #769, opened 2022-11-02
Make loading plugins which are packages work better

rhelmot:


ltfish:
Now that our tests work better, can we have a test case?

ltfish (approved):

Issue #783, opened 2022-11-08
Log Widget is not showing all log messages


zwimer:
### Description

~~This is presumed to be a consequence of merging: https://github.com/angr/angr/pull/3590~~

This is because multiprocess workers cannot properly write to the main process' log.

### Environment

Master: https://github.com/angr/angr-management/commit/0a28d8c62eac78302e6e085c0c9cb0ee67a64136

### Additional context

The likely responsible PR converted the 'corruption' log handler CuteHandler to a standard log handler that uses a custom formatter. angr-management might have had custom code to prevent the corruption / hook logging- we should start looking there.

zwimer:
This statement has not been verified, it is simply what we believe is likely:
> This is presumed to be a consequence of merging: https://github.com/angr/angr/pull/3590

zwimer:
Update: It is likely not related to the recent logging-centered PRs in `angr`.

zwimer:
To accomplish this, a PR must:
- [ ] Allow transporting of log messages from subprocesses to the main process logger
- [ ] Each subprocess must install this 'transit' logger on creation

PR #786, opened 2022-11-09
Allow angr-management's log widget to capture subprocess logging

zwimer:
Fixes: https://github.com/angr/angr-management/issues/783

As I have an M1, my testing ability is limited.

Requires:
- [x] https://github.com/angr/angr/pull/3613

### DIFF
--- a/angrmanagement/data/instance.py
+++ b/angrmanagement/data/instance.py
@@ -18,7 +18,7 @@ from .analysis_options import AnalysesConfiguration, CFGAnalysisConfiguration, F
 from .jobs import VariableRecoveryJob, PrototypeFindingJob, CodeTaggingJob, FlirtSignatureRecognitionJob, \
     CFGGenerationJob
 from .object_container import ObjectContainer
-from .log import LogRecord, LogDumpHandler
+from .log import LogRecord, initialize
 from ..logic import GlobalInfo
 from ..logic.threads import gui_thread_schedule_async, gui_thread_schedule
 from ..logic.debugger import DebuggerListManager, DebuggerManager
@@ -95,10 +95,8 @@ class Instance:
         self._label_rename_callback = None  # type: Union[None, Callable[[int, str], None]]      #  (addr, new_name)
         self._set_comment_callback = None  # type: Union[None, Callable[[int, str], None]]       #  (addr, comment_text)
 
-        self._logging_handler = LogDumpHandler(self)
-
-        # Register a root logger
-        logging.root.handlers.insert(0, self._logging_handler)
+        # Setup logging
+        initialize(self)
 
         self.cfg_args = None
         self.variable_recovery_args = None
--- a/angrmanagement/data/instance.py
+++ b/angrmanagement/data/instance.py
@@ -79,7 +79,7 @@ class Instance:
                                 lambda: [PlainTextProtocol, BackslashTextProtocol],
                                 List[Type[ProtocolInteractor]],
                                 'Available interaction protocols')
-        self.register_container('log', lambda: [], List[LogRecord], 'Saved log messages')
+        self.register_container('log', lambda: [], List[LogRecord], 'Saved log messages', async_=True)
         self.register_container('current_trace', lambda: None, Type[Trace], 'Currently selected trace')
         self.register_container('traces', lambda: [], List[Trace], 'Global traces list')
 
@@ -170,7 +170,7 @@ class Instance:
     # Public methods
     #
 
-    def register_container(self, name, default_val_func, ty, description):
+    def register_container(self, name, default_val_func, ty, description, *, async_: bool = False):
         if name in self.extra_containers:
             cur_ty = self._container_defaults[name][1]
             if ty != cur_ty:
@@ -178,7 +178,7 @@ class Instance:
 
         else:
             self._container_defaults[name] = (default_val_func, ty)
-            self.extra_containers[name] = ObjectContainer(default_val_func(), description)
+            self.extra_containers[name] = ObjectContainer(default_val_func(), description, async_=async_)
 
     def initialize(self, initialized=False, **kwargs):  # pylint:disable=unused-argument
         if self.project.am_none:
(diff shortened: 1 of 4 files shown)


zwimer:
@mborgerson I should note that this change only synchronized the EventSentinel parts. That is, the following line is *not* synchronized between processes:
https://github.com/angr/angr-management/blob/f058fb1e2a5910445f2be4f067e6f2cf30ed24ed/angrmanagement/data/log.py#L29
Only this is:
https://github.com/angr/angr-management/blob/f058fb1e2a5910445f2be4f067e6f2cf30ed24ed/angrmanagement/data/log.py#L30

rhelmot:
Two main issues I have with this, which should probably be solved at the same time:

- One thread per sentinel seems like a lot
- sentinels do not have unbounded lifetimes, they should have their threads (or references in whatever one thread handles the dispatch) cleaned up on `__del__`.

zwimer:
@rhelmot Their daemonic thread should terminate not long after their lifetime, the loop body takes in a weak reference to self for this reason. Or were you referring to making a `join()` call on the EventSentinel `__del__`?

Regarding one thread per sentinel; it shouldn't be an issue since right now it's only used for logging, but if you are trying to future-proof, what I can do is make a single dispatch thread that has a `dict` of `weakptr.Proxy` objects mapped to `EventSentinel`; this thread would basically preform the same actions as the present one except it would remove dead proxies from the set it is monitoring rather than die along with a proxy; and `time.sleep(.01)` or something if there are no items in the set. Thoughts?

zwimer:
After further investigation, I've noticed this doesn't fix the problem. It does allow async / interprocess EventSentinels, but it seems like child processes do not have the logging handlers set themselves.

If we do want to keep the async abilities introduce by this PR I can re-create this PR, but it will not fundamentally fix: https://github.com/angr/angr-management/issues/783

To fix this issue, I see the three options:

#### Manually adding handlers each time

Each `multiprocessing.Process` would have to add a logging manually (or via some function); perhaps a `logging.QueueHandler` wrapping a `multiprocessing.Queue`. The downside of this

The downside of this is that we can only do this for the `multiprocessing.Process` that `angr-mangagement` directly spawns; that is to say the `multiprocessing.Process` calls in `angr` itself and below would not install this logging module, thus their messages would be missed.

#### Remapping stdout / stderr

If we remap `stdout` (and perhaps `stderr`) to a pipe, then the processes' default handlers will simply write their log messages to a pipe. We could then change the base process' logging handler to output to the 'new' stdout/stderr. Maybe something like:
```python
# Remap stderr
real_stderr = sys.stderr
log_pipe_read, log_pipe_write = os.pipe()
sys.stderr = log_pipe_write
# Allow main process to log to stderr
logging.addHandler(logging.StreamHandler(real_stderr)
# New children processes will now log to: log_pip_write
```

The downside here is that we would only be able to get text from these messages; so we would have to strip out the ansi escape codes `angr` and parse the log message; probably `msg.split("|", maxsplit=4)`.

#### Forcing every process to load a given file 

Is there some sort of `LD_PRELOAD` equivalent for `multiprocessing.Process`? I know there is for pools, but just for `Process`? I can think of a few hacky ways to do this, but none that are actually 'good'. This would be the ideal solution, having each process just add the new logging handler themselves on init, but I don't know how to accomplish this.

rhelmot:
The equivalent of LD_PRELOAD for multiprocessing.Process is that it will attempt to re-import the `__main__` module in each process. You can detect that it isn't the main process because `__name__ != '__main__'`. This is the correct solution I believe.

zwimer:
@rhelmot It's `__mp_main__`, but that only runs on files the process loads; but subprocesses won't necessarily load the code with this.

To test this I inserted file writes in the `bin/angrmanagement` file along with the top level `__init__.py` and `__main__.py` files that would only trigger on `__mp_main__`. I saw no writes when launching angr-management then loading `/bin/false`; despite noticing angr-management said it was spawning workers (this should have spawned angr processes for variable recovery I think).

Is there perhaps an `LD_PRELOAD`-like hook for `multiprocessing.Process` that uses `POpen`'s `preexec_fn` or something? Since I don't know if we can rely on any given python file to import for certain. At least as far as I'm aware. I also checked for all `__name__` to see if it was something like `angrmanagement.__mp_main__` but only the main process ever invoked this code so no `multiprocessing.Process`' ever ran this code.

zwimer:
Unless we manually add a callback to every `multiprocessing.Process` target functions in both `angr`, `angr-management`, and others.

rhelmot:
Looked into it a bit more. I believe this is a problem with the `spawn` vs `fork` start methods. spawn will re-import `__mp_main__` but fork will not. Since we are using fork on linux there is no way to hook process start.

I believe the only solution is for each analysis which can be multiprocessed to manually manage its own logs. We can provide a tool in angr (`angr.utils.logging.log_to_parent()`) to make this significantly easier.

twizmwazin:
@rhelmot Python has a way to set pre- and post-fork callbacks https://docs.python.org/3/library/os.html#os.register_at_fork

zwimer:
I was testing on macOS

zwimer:
In this case, I will just have angr hold some contextual object meant to hold IPC objects which it will pass to each `multiprocessing.Process` invocation as an arg. For example `ipc = { "logging": multiprocessing.Queue() }` (or as a named tuple or something). This could be private; with perhaps public methods for things like `init_process()` which might hook up logging or other IPC related things that should remain across all subprocesses, as @rhelmot seconded earlier.

zwimer:
CI is only failing for linting for push, not pull request, because the linting is complaining about items in angr not existing since it isn't checking out the PR branch of angr this PR depends on (the pull request version does though)

zwimer:
@mborgerson This should fix the issue now that the related PR has been merged into angr.

zwimer:
I cannot test this for Linux. It's only been tested on macos

zwimer:
Tested on both macOS and ubuntu; this should cover both fork and spawn as well.

mborgerson:
Thanks!

PR #789, opened 2022-11-09
Add version option to angr management CLI

zwimer:


angrmanagement/__main__.py
```diff
@@ -198,7 +203,9 @@ def main():
     import argparse
     signal.signal(signal.SIGINT, signal.SIG_DFL)
 
-    parser = argparse.ArgumentParser(description="angr management")
+    prog: str = name.replace(" ", "-")
+    parser = argparse.ArgumentParser(prog=prog, description=name)
+    parser.add_argument('--version', action='version', version=f"{prog} {__version__}")
```
twizmwazin, line 208: Can we add -v to this?

angrmanagement/data/object_container.py
```diff
@@ -29,6 +29,12 @@ def am_event(self, **kwargs):
 
 
 class ObjectContainer(EventSentinel):
+    """
```
twizmwazin, line 32: This docstring looks like it is not supposed to be part of this PR

angrmanagement/__main__.py
```diff
@@ -9,6 +9,8 @@
 import platform
 import signal
 
+from . import __version__
```
twizmwazin, line 12: I personally prefer `from angrmanagement import __version__` instead of the relative import.

angrmanagement/data/object_container.py
zwimer: It was for ensuring CI passed linting and I have no issues adding documentation to code anywhere.

angrmanagement/__main__.py
zwimer: I have no issues with relative imports from the same directory. I.e. one dot is fine with me. If you dislike it you can change it but it doesn't play as well with tooling I've found.

PR #792, opened 2022-11-19
Fix app title on mac

zwimer:
Upon merging this we probably want to open an issue to fix the application name itself on a mac as this only fixes the title. A PR which fixes the application name will likely make this code redundant.

To make this work for the dock: If we used `CFBundleDisplayName` instead of `CFBundleName` this would set the name in the dock as well, but that string is loaded into the dock and such before the app has a chance to change it, so changing it afterwards wouldn't work (well, it might for future things that query for the name, if such things exist).

It might be dynamically editable via Cocoa; some mac apps do dynamically edit their dock entries.

### DIFF
--- a/angrmanagement/__main__.py
+++ b/angrmanagement/__main__.py
@@ -8,7 +8,6 @@ import time
 import warnings
 import platform
 import signal
-import pathlib
 
 if sys.platform.startswith("darwin"):
     from Foundation import NSBundle
@@ -149,7 +148,7 @@ def start_management(filepath=None, use_daemon=None, profiling=False):
         if bundle:
             app_info = bundle.localizedInfoDictionary() or bundle.infoDictionary()
             if app_info:
-                app_info['CFBundleName'] = pathlib.Path(name).stem
+                app_info['CFBundleName'] = name
 
     app = QApplication(sys.argv)
     app.setApplicationDisplayName(name)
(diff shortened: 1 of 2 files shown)


zwimer:
Tested on an M1; verified `pip install` handles the platform specific dependencies (as per a pep) on Fedora.

zwimer:
@twizmwazin If you feel like diving into Cocoa a bit, you might be able to edit the dock name with similar code to this: https://github.com/tauri-apps/tauri/issues/2985

angrmanagement/__main__.py
```diff
@@ -9,7 +9,11 @@
 import platform
 import signal
 
-from . import __version__
+if sys.platform.startswith("darwin"):
+    from Foundation import NSBundle  # pylint: disable=import-error
+
+from . import __version__  # pylint: disable=wrong-import-position
```
twizmwazin, line 15: Let's put this above the conditional import so the linter is happy

zwimer: I don't mind, I ignored the linter since the convention was 'this package' import last, but I can rearrange them.

twizmwazin: I always put conditional imports after other imports, they're outside of the standard rules since they have logic associated with them.

Issue #797, opened 2022-11-21
Exception on close


mborgerson:
### Description

```
$ angr-management /bin/true
0.00s - Debugger warning: It seems that frozen modules are being used, which may
0.00s - make the debugger miss breakpoints. Please pass -Xfrozen_modules=off
0.00s - to python to disable frozen modules.
0.00s - Note: Debugging will proceed. Set PYDEVD_DISABLE_FILE_VALIDATION=1 to disable this validation.
WARNING  | 2022-11-21 14:34:51,884 | cle.loader     | The main binary is a position-independent executable. It is being loaded with a base address of 0x400000.
WARNING  | 2022-11-21 14:34:53,445 | cle.backends.externs | Symbol was allocated without a known size; emulation may fail if it is used non-opaquely: program_invocation_name
WARNING  | 2022-11-21 14:34:53,446 | cle.backends.externs | Symbol was allocated without a known size; emulation may fail if it is used non-opaquely: program_invocation_short_name
WARNING  | 2022-11-21 14:34:53,446 | cle.loader     | The main binary is a position-independent executable. It is being loaded with a base address of 0x400000.
WARNING  | 2022-11-21 14:34:53,447 | cle.loader     | For more information about "Symbol was allocated without a known size", see https://docs.angr.io/extending-angr/environment#simdata
WARNING  | 2022-11-21 14:34:56,452 | angr.analyses.reaching_definitions.function_handler | Please implement the indirect function handler with your own logic.
WARNING  | 2022-11-21 14:34:56,592 | angr.analyses.reaching_definitions.function_handler | No handler for external function __libc_start_main(), falling back to generic handler
Exception in thread Thread-1 (_monitor):
Traceback (most recent call last):
  File "/usr/lib/python3.11/threading.py", line 1038, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.11/threading.py", line 975, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/lib/python3.11/logging/handlers.py", line 1572, in _monitor
    record = self.dequeue(True)
             ^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/logging/handlers.py", line 1521, in dequeue
    return self.queue.get(block)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/multiprocessing/queues.py", line 103, in get
    res = self._recv_bytes()
          ^^^^^^^^^^^^^^^^^^
[...]

PR #799, opened 2022-11-22
Cleanup logger on exit

zwimer:
Fixes: https://github.com/angr/angr-management/issues/797


zwimer:
Tested on ubuntu; loaded `/bin/true` and saw no crash on close.

Issue #805, opened 2022-11-30
No 'current' theme


zwimer:
### Description

When loading a theme in angr-management, there are 3 options: Light, Dark, and Current. Selecting `Current` can cause:
```python
Traceback (most recent call last):
  File "/Users/zwimer/.virtualenvs/am-dev/lib/python3.10/site-packages/angrmanagement/ui/dialogs/preferences.py", line 151, in _on_load_scheme_clicked
    self._load_color_scheme(self._schemes_combo.currentText())
  File "/Users/zwimer/.virtualenvs/am-dev/lib/python3.10/site-packages/angrmanagement/ui/dialogs/preferences.py", line 146, in _load_color_scheme
    for prop, value in COLOR_SCHEMES[name].items():
KeyError: 'Current'
```

### Steps to reproduce the bug

```bash
rm -rf ~/.config/angr-management
angr-management
```
Open preferences
"Themes and Colors"
Beside "Load Theme" select Current
Click "Load"

### Environment

https://github.com/angr/angr-management/commit/41b574ed070e8112b1cf5e61e43e81286d569c86

zwimer:
@mborgerson What is "Current" as a theme choice supposed to do? A no-op?

zwimer:
Is it supposed to be the System theme? If so, I think we should rename this to "System".

mborgerson:
'Current' is supposed to mean whatever is currently configured. e.g. if the user loads the Dark theme, Dark theme values are copied to Current

zwimer:
What is the rational for having "Current" when the loaded theme is already in the combobox selector and should be auto-selected when the preference tab opens?

zwimer:
I'll remove the `Current` for now, this will is done in #806

Issue #780, opened 2022-11-07
Allow theme based on system preferences


mborgerson:
### Description

Many apps are themed based on system settings. It would be nice to support using the user's system-wide color scheme (dark / light) on startup.

zwimer:
Seems useful: https://github.com/albertosottile/darkdetect

zwimer:
They have a listener class as well that can be a thread target.

PR #806, opened 2022-12-01
Add system theme support

zwimer:
Implements: https://github.com/angr/angr-management/issues/780
Fixes: https://github.com/angr/angr-management/issues/805

### DIFF
--- a/angrmanagement/utils/track_system_theme.py
+++ b/angrmanagement/utils/track_system_theme.py
@@ -41,6 +41,35 @@ class TrackSystemTheme:
     _system: str = "System"
 
     #
+    # Private methods
+    #
+
+    def __init__(self, parent: Optional[QObject], *, _caller=None):
+        """
+        This method is not public
+        """
+        if _caller != self.create:  # pylint: disable=comparison-with-callable
+            raise RuntimeError("Use .create(parent) or .get(); this is a singleton")
+        # Init
+        self._lock = Lock()
+        self._parent = parent
+        self._underlying: str = darkdetect.theme()
+        self._enabled: bool = False
+        self._listener: Optional[_QListener] = None
+        self._thread: Optional[QThread] = None
+
+    def _set_theme(self, theme: str, *, force: bool = False):
+        """
+        Set the underlying theme according to the system theme if needed
+        """
+        if force or theme != self._underlying:
+            self._underlying = theme
+            Conf.theme_name = self._underlying
+            for prop, value in COLOR_SCHEMES[theme].items():
+                setattr(Conf, prop, value)
+            gui_thread_schedule_async(refresh_theme)
+
+    #
     # Public methods
     #
 
@@ -98,33 +127,4 @@ class TrackSystemTheme:
         if self.enabled():
             self._set_theme(darkdetect.theme(), force=True)
         else:
-            gui_thread_schedule_async(refresh_theme)
-
-    #
-    # Private methods
-    #
-
-    def __init__(self, parent: Optional[QObject], *, _caller=None):
-        """
-        This method is not public
-        """
-        if _caller != self.create:  # pylint: disable=comparison-with-callable
-            raise RuntimeError("Use .create(parent) or .get(); this is a singleton")
-        # Init
-        self._lock = Lock()
-        self._parent = parent
-        self._underlying: str = darkdetect.theme()
-        self._enabled: bool = False
-        self._listener: Optional[_QListener] = None
-        self._thread: Optional[QThread] = None
-
-    def _set_theme(self, theme: str, *, force: bool = False):
-        """
-        Set the underlying theme according to the system theme if needed
-        """
-        if force or theme != self._underlying:
-            self._underlying = theme
-            Conf.theme_name = self._underlying
-            for prop, value in COLOR_SCHEMES[theme].items():
-                setattr(Conf, prop, value)
             gui_thread_schedule_async(refresh_theme)


zwimer:
A better way to do this would be to add an observer with a callback, rather than polling. I'll investigate this.

zwimer:
I made a PR on dark detect to add a listener for macoOS: https://github.com/albertosottile/darkdetect/pull/30

Once it is merged and updated on pypi, I can update the PR to use the listeners instead of doing polling.

zwimer:
In order to avoid leaving potentially orphaned subprocesses around, I've filed this issue: https://github.com/albertosottile/darkdetect/issues/31

zwimer:
Once https://github.com/albertosottile/darkdetect/pull/32 is merged I'll update this PR; hopefully a new PyPi release will be out soon at that point.

zwimer:
CI will continue to fail until the `darkdetect` PR is merged and a new pypi release is made.

zwimer:
The test failures are also failing on master.

zwimer:
Upon merging, #780 and #805 should both be marked as complete.

angrmanagement/ui/dialogs/preferences.py
```diff
@@ -172,11 +190,16 @@ def _on_load_scheme_clicked(self):
         self._load_color_scheme(self._schemes_combo.currentText())
         self.save_config()
 
+    def revert_unsaved(self):
```
mborgerson, line 193: Identical to base class implemantation

angrmanagement/ui/dialogs/preferences.py
```diff
@@ -235,6 +259,9 @@ def save_config(self):
         for i in self._font_options:
             i.update()
 
+    def revert_unsaved(self):
```
mborgerson, line 262: Identical to base class implemantation

angrmanagement/ui/dialogs/preferences.py
```diff
@@ -203,7 +226,8 @@ def _init_widgets(self):
         fmt: str = Conf.log_timestamp_format
         ts = datetime.now()
         # pylint: disable=use-sequence-for-iteration
-        self._fmt_map = bidict({ts.strftime(i): i for i in {fmt, "%X", "%c"}})  # set also dedups
+        self._fmt_map = bidict({ts.strftime(i): i for i in ("%X", "%c")})
```
mborgerson, line 229: Should probably be in a different PR

angrmanagement/utils/track_system_theme.py
```diff
@@ -0,0 +1,130 @@
...
+        if self.enabled():
+            self._set_theme(darkdetect.theme(), force=True)
+        else:
+            gui_thread_schedule_async(refresh_theme)
+
+    #
+    # Private methods
+    #
+
+    def __init__(self, parent: Optional[QObject], *, _caller=None):
```
mborgerson, line 107: nit: Prefer constructor to be at the top of class definition, even if direct instantiation is disallowed

setup.cfg
```diff
@@ -29,6 +29,7 @@ install_requires =
     qtterm
     requests[socks]
     tomlkit
+    darkdetect-angr[macos-listener]
```
mborgerson, line 32: * It's unfortunate that we have to maintain a forked release of darkdetect now, but I appreicate the work on it
* Did Windows/Linux get tested?

angrmanagement/ui/dialogs/preferences.py
zwimer: This is actually a bug fix that prevents duplicate keys from appearing in the initial dictionary, which previously might remove the presently installed format from the dict. That's why it's now inserted after the fact to ensure that it exists; this has the side effect that other formats in that initial set that map to the exact same timestamp when formatted will not appear as default options, but that's fine. (I.e. if "%A" and "%B" are different format strings that yield the same output, this `bidict` cannot handle it so the presently installed one is preferred over any 'default options')

setup.cfg
zwimer: The majority of the fork has a PR into the main repo, so once that is merged we might not need to hopefully. TBD. I've tested the PR'd version on all three OS's, though less throughly on Windows as I lack said OS. The version we are using is mostly the same so it ought to work. Feel free to test them yourself but the relevant logic that distinguished between OSes wasn't really the code that changed between the PR'd version and the hardfork version.

angrmanagement/ui/dialogs/preferences.py
zwimer: Lacking this fix, an exception will be raised on launch.

angrmanagement/ui/dialogs/preferences.py
zwimer: The base class raises `NotImpelementedError`.

angrmanagement/ui/dialogs/preferences.py
zwimer: The base class raises `NotImpelementedError`.

angrmanagement/utils/track_system_theme.py
zwimer: I can reorder it.

angrmanagement/ui/dialogs/preferences.py
mborgerson: Oh, I guess I was looking at the wrong thing 

Why not have the default implementation do nothing instead of raising an error? Or just drop the method because there's not an implementation that does anything?

setup.cfg
mborgerson: > I've tested the PR'd version on all three OS's

But not the forked version with extra changes?

> Feel free to test them yourself

I don't plan to test this PR. Will you please test it

zwimer: The extra changes are copyright, README, and `__version__`, and `pyproject.toml` to change the package name. Not really the code.

zwimer: If you are curious, here is the diff stats:
```
 LICENSE                     |  4 ++--
 README.md                   |  8 ++++----
 darkdetect/__init__.py      |  4 ++--
 darkdetect/__main__.py      |  2 +-
 darkdetect/_dummy.py        |  2 +-
 darkdetect/_linux_detect.py |  2 +-
 docs/api.md                 |  2 +-
 docs/examples.md            |  2 +-
 pyproject.toml              | 10 +++++-----
 ```
 Note the 2 line changes are copyright comments. I don't have access to a Windows device at the moment so I can't really test these changes easily, but they aren't really code changes just the changes necessary to hard-fork a project (changing the name on pypi, copyright, documentation, and version).

twizmwazin: Github actions supports windows, why not write some test cases?

zwimer: I do not know how to automate GUI test cases on windows changing system theme and detecting of the colors properly all changed over. Feel free. The actions to test this would be launch app, open preferences, select 'Track System Theme', then toggle the Windows System Theme and visually ensure it tracks the theme.

twizmwazin: You don't need to "visually ensure" it, you just need to make sure after the theme is changed the correct behavior occurs in the library. I'm sure applescript and powershell each have ways to adjust the system theme.

twizmwazin: That's on the angr management side, but not necessary for the library itself which is is the subject of this thread.

zwimer: I'll happily take a PR for them here: https://github.com/zwimer/darkdetect but I don't have plans to write tests myself for them at the moment. If you do PR them, I'll also merge them into the branch that is PR-ing into the original repo https://github.com/albertosottile/darkdetect so as to avoid actively diverging the hard fork from the original more than necessary; my plan is to get rid of the fork once the PR https://github.com/albertosottile/darkdetect/pull/32 in the original is merged.

Right now that PR has been tested on all three OSes and my fork's master branch doesn't meaningfully diverge from said code. I know the maintainer desired to do more thorough testing before merging the PR though, so I'm sure the test cases would be appreciated by all parties.

Issue #810, opened 2022-12-01
Better Keyboard Shortcut for Preferences


zwimer:
### Description

To open preferences, right now on mac it is Command-`p`. This is 'print' on every other app; in windows ctrl-p is also print. We should change this.

We should change it to:
For macOS: Command-`,`
For Windows: TBD
For Linux: TBD

zwimer:
@twizmwazin What does Windows normally use as a shortcut for preferences? On macOS, Command-`,` on most any app will open that app's preferences; it's rather standard like Command-`q` for quit. What is the 'preferences' shortcut for Windows? I assume the same for Linux?

twizmwazin:
Pyside/Qt has [this](https://doc.qt.io/qtforpython/PySide6/QtGui/QKeySequence.html#standard-shortcuts) nifty tables that shows all of the common built-in keyboard shortcut. Only macOS has a standard preferences key.

PR #811, opened 2022-12-01
Use Ctrl-, instead of Ctrl-p as a keyboard shortcut for opening preferences

zwimer:
Implements: https://github.com/angr/angr-management/issues/810
`,` is standard for mac, there seems to be no standard for windows so I'm keeping comma to be consistent. `p` is for print on all OSes

### DIFF
--- a/angrmanagement/ui/menus/file_menu.py
+++ b/angrmanagement/ui/menus/file_menu.py
@@ -40,7 +40,7 @@ class FileMenu(Menu):
             MenuSeparator(),
             MenuEntry('Load a new &trace...', main_window.load_trace),
             MenuSeparator(),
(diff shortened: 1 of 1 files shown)


twizmwazin:
I poked around at a few applications on windows, and some applications implement CTRL-, as preferences, other applications don't appear to register a shortcut at all. VSCode, Windows Terminal, Discord, Slack all use CTRL-, on Windows.

zwimer:
@twizmwazin Can you let me know how this works for you on your Windows machine?

twizmwazin:
You had it correct before. Since there doesn't seem to be a convention on windows, applications choosing to implement a keybind carry over the equivalent binding from macos, using "Control" and "Comma", not "Control" and "Dash".

twizmwazin (approved):

Issue #745, opened 2022-10-20
Make log message timestamp more human-friendly


mborgerson:
### Description

Today the timestamp is formatted in seconds-since-epoch style, which is not very friendly.

![Screenshot from 2022-10-20 12-37-19](https://user-images.githubusercontent.com/8210/197041775-35b793d7-89dd-49a6-8a7b-0a209cd39ac2.png)

Here are some suggestions to explore
* Configurable format (yyyy-MM-dd hh:mm:ss.xxx style) with option to enable or disable parts of the timestamp
* Option to show timestamps relative to start of program

### Alternatives

N/A

PR #813, opened 2022-12-01
Adjustable log timestamp format implemented

zwimer:
Allows a user-adjustable timestamp as per: https://github.com/angr/angr-management/issues/745

Note: This does not currently support 'time since program start'; but adding support should be relatively easy


mborgerson:
Thanks!

Issue #817, opened 2022-12-02
Log Timestamp Presets


zwimer:
### Description

The log timestamp setting should offer presets

PR #821, opened 2022-12-06
Provide default options for log format

zwimer:
Implements: https://github.com/angr/angr-management/issues/817

### DIFF
--- a/angrmanagement/__main__.py
+++ b/angrmanagement/__main__.py
@@ -93,7 +93,7 @@ def start_management(filepath=None, use_daemon=None, profiling=False):
             info["CFBundleName"] = name
         except Exception as e:  # pylint: disable=broad-except
             # This happens before logging is setup so use stderr
-            print(f"{type(e).__name__}: {e}", file=sys.stderr)
+            print(f"Failed to set App name! {type(e).__name__}: {e}", file=sys.stderr)
 
     app = QApplication(sys.argv)
     app.setApplicationDisplayName(name)
--- a/angrmanagement/ui/dialogs/preferences.py
+++ b/angrmanagement/ui/dialogs/preferences.py
@@ -1,3 +1,6 @@
+from datetime import datetime
+from bidict import bidict
+
 from PySide6.QtGui import QColor
 from PySide6.QtWidgets import (
     QDialog,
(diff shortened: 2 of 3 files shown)


angrmanagement/ui/dialogs/preferences.py
```diff
@@ -179,9 +179,13 @@ def _init_widgets(self):
         log_format_lbl = QLabel("Log datetime Format String:")
         log_format_lbl.setSizePolicy(QSizePolicy(QSizePolicy.Fixed, QSizePolicy.Fixed))
         log_format_layout.addWidget(log_format_lbl)
-        self.log_format_entry = QLineEdit(self)
-        self.log_format_entry.setClearButtonEnabled(True)
-        self.log_format_entry.setText(Conf.log_timestamp_format)
+
+        self.log_format_entry = QComboBox(self)
+        fmt: str = Conf.log_timestamp_format
+        for i in {fmt, "%X", "%c"}:  # pylint: disable=use-sequence-for-iteration
```
twizmwazin, line 185: Why are you using a set and then disabling the warning? Why not just make this a list?

zwimer: Deduplication

angrmanagement/ui/dialogs/preferences.py
```diff
@@ -179,9 +179,13 @@ def _init_widgets(self):
...
+            self.log_format_entry.addItem(i)
```
mborgerson, line 186: It would be nice to show a format example because I cannot remember what `%X` and `%c` mean in this context

zwimer: Implemented

zwimer:
An example:
image

mborgerson:
@zwimer: Please rebase

angrmanagement/__main__.py
```diff
@@ -90,7 +90,7 @@ def drawContents(self, painter):
             info["CFBundleName"] = name
         except Exception as e:  # pylint: disable=broad-except
             # This happens before logging is setup so use stderr
-            print(f"{type(e).__name__}: {e}", file=sys.stderr)
+            print(f"Failed to set App name! {type(e).__name__}: {e}", file=sys.stderr)
```
twizmwazin, line 93: Wasn't this a different PR?

angrmanagement/ui/dialogs/preferences.py
```diff
@@ -1,3 +1,6 @@
+from datetime import datetime
+from bidict import bidict
```
twizmwazin, line 2: This is third-party

angrmanagement/ui/dialogs/preferences.py
```diff
@@ -177,9 +180,17 @@ def _init_widgets(self):
...
         log_format_layout.addWidget(log_format_lbl)
-        self.log_format_entry = QLineEdit(self)
-        self.log_format_entry.setClearButtonEnabled(True)
-        self.log_format_entry.setText(Conf.log_timestamp_format)
+
+        self.log_format_entry = QComboBox(self)
+        fmt: str = Conf.log_timestamp_format
+        ts = datetime.now()
+        # pylint: disable=use-sequence-for-iteration
+        self._fmt_map = bidict({ ts.strftime(i): i for i in {fmt, "%X", "%c"} })
```
twizmwazin, line 188: I still think using a set to implicitly de-duplicate values is not a great pattern

angrmanagement/__main__.py
zwimer: It's just cleaning up the error message. I had to do it for debugging something earlier, left it in because it's a good change that'll save someone else debugging later should this ever happen to them for some reason.

angrmanagement/ui/dialogs/preferences.py
zwimer: It's what a set is for in this case; you care about the items being unique, not the order or anything; that's the set data structure use-case.

angrmanagement/ui/dialogs/preferences.py
zwimer: It is

Issue #519, opened 2021-11-04
View switching should use `Alt`, instead of `Ctrl` key


mborgerson:
Currently using <kbd>Ctr</kbd>+<kbd>[0-9]</kbd> to switch tabs. Web browsers, Ida use <kbd>Alt</kbd>+<kbd>[0-9]</kbd>. angr-management should align with this convention.

PR #830, opened 2022-12-07
Prefer Alt+Tab to Ctrl+Tab for view switching

zwimer:
Implements: https://github.com/angr/angr-management/issues/519

### DIFF
--- a/angrmanagement/ui/menus/view_menu.py
+++ b/angrmanagement/ui/menus/view_menu.py
@@ -75,9 +75,9 @@ class ViewMenu(Menu):
         self.entries.extend([
             ToolbarMenu(main_window),
             MenuSeparator(),
-            MenuEntry('Next Tab', main_window.workspace.view_manager.next_tab, shortcut=QKeySequence("Alt+Tab")),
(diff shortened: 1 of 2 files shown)


angrmanagement/ui/menus/view_menu.py
```diff
@@ -75,9 +75,9 @@ def __init__(self, main_window):
         self.entries.extend([
             ToolbarMenu(main_window),
             MenuSeparator(),
-            MenuEntry('Next Tab', main_window.workspace.view_manager.next_tab, shortcut=QKeySequence("Ctrl+Tab")),
+            MenuEntry('Next Tab', main_window.workspace.view_manager.next_tab, shortcut=QKeySequence("Alt+Tab")),
```
mborgerson, line 78: We should avoid re-binding Alt-Tab because it's already used for global window switching

zwimer: Would you prefer to have this one line reverted, just having `Alt[0-9]` remain? Edit the global switching to something else? Or Close this PR @mborgerson ?

Issue #653, opened 2022-05-04
'Save angr database' menu options should be disabled if there is no active project


mborgerson:

zwimer:
This also causes some level of freezing. So I'm adding bug.

PR #832, opened 2022-12-07
Save greyed out when no project loaded; `Menu` updates `QAction` to saved state on `set_action`

zwimer:
Fixes: https://github.com/angr/angr-management/issues/653

### DIFF
--- a/angrmanagement/data/instance.py
+++ b/angrmanagement/data/instance.py
@@ -37,7 +37,7 @@ class Instance:
     """
     An object to give access to normal angr project objects like a Project, CFG, and other analyses.
     """
-    project: Union[angr.Project, ObjectContainer]
+    project: ObjectContainer
     cfg: Union[angr.analyses.cfg.CFGBase, ObjectContainer]
     cfb: Union[angr.analyses.cfg.CFBlanket, ObjectContainer]
     log: Union[List[LogRecord], ObjectContainer]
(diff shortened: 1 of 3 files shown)


angrmanagement/data/instance.py
```diff
@@ -37,7 +37,7 @@ class Instance:
     """
     An object to give access to normal angr project objects like a Project, CFG, and other analyses.
     """
-    project: Union[angr.Project, ObjectContainer]
+    project: ObjectContainer
```
mborgerson, line 40: Presumably the annotations were like this for code completion. We should be consistent in style though; that is, either switch all of these lines over to annotate only ObjectContainer, or do the Union with the real type for completion. I'll defer to @ltfish on it

angrmanagement/ui/menus/file_menu.py
```diff
@@ -45,6 +52,11 @@ def __init__(self, main_window):
             MenuEntry('E&xit', main_window.quit),
         ])
 
+    def _edit_save(self, **_):
+        enable: bool = not self._project.am_none
+        for i in self._save_entries:
+            (i.enable if enable else i.disable)()
```
mborgerson, line 58: nit: Simpler to just pass the bool in to a setEnabled function

zwimer: There is no `setEnabled` function; we are editing from the custom class which exposes only these two functions- correct me if I am wrong.

zwimer: https://github.com/angr/angr-management/blob/6c70c966e52ffc99e5d609c3e5261888f4bc5a3c/angrmanagement/ui/menus/menu.py#L5

PR #837, opened 2022-12-11
Make multiprocessing work fine on Windows PyInstaller builds.

### LINKED ISSUE #836: Hang/crash when switching functions using pyinstaller build
### Description

When switching between functions after loading a binary, the application hangs. This appears to be specific to the pyinstaller build.

1. Open angr management onefile build
2. Load binary (I used `/usr/bin/true` from ubuntu 20.04)
3. Select a different function other than `main` (I tried _start)

### Environment

Windows 11 64 bit

### Additional context

_No response_

ltfish:
Close #836.

### DIFF
--- a/angrmanagement/__main__.py
+++ b/angrmanagement/__main__.py
@@ -1,5 +1,6 @@
 # pylint:disable=import-outside-toplevel,unused-import,no-member
 import asyncio
+import multiprocessing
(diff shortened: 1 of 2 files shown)


twizmwazin:
Yep this fixes that issue on windows!

Issue #835, opened 2022-12-08
"Load a new docker target" is shown in file menu even if archr is not installed


twizmwazin:
### Description

In environments without archr, the option "Load a new docker target" is shown in the file menu. This also applies to the pyinstaller bundled version of the program when archr is not bundled.

zwimer:
Actually, is this a bug? Right now it informs the user that docker has to be installed to use this feature rather than just not existing. This was an intentional UI decision iirc. Idk if we want to keep it for certain, but I think this is not a bug for sure.

zwimer:
The same is true with the docker icon in the top bar of the UI.

zwimer:
If we want this, we should make `archr` an optional dependency. So users can do `pip install angr-management[docker]` rather than telling them to pip install arch and keystone-engine with `--no-binary` standalone.

twizmwazin:
We removed the docker button if archr is not installed. pip install instructions aren't valid in pyinstaller builds so the current error message is simply incorrect.

zwimer:
Related, but we have other optional dependencies we may want to do this for, like `nclib`, and `slacrs`.

zwimer:
I'll move this discussion offline and fix this bug.

Issue #844, opened 2022-12-13
Get rid of soft-dependencies (excluding non-core plugins)


zwimer:
### Description

Right now `angr management` has multiple soft dependencies:

In `interaction_view.py`, for example, we soft require: `archr`, `keystone-engine`, `nclib`, and `slacrs` (though, this view seems useless without any of these installed).

`pyinstaller` gives us a frozen python environment, which means these features are simply not-usable for many of our users. I would (and still do) suggest moving something like the `interactions_view` into a plugin, but if it is a default plugin we ship, it will simply be unusable for users of the pyinstaller build.

With that in mind, I think we should do one of the following:
1. Get rid of these soft dependencies by moving the features into non-default plugins.
2. Make these soft dependencies required.
3. Move any code requiring a soft dependency into a (default) plugin, adding the python packages as `extras_require` in our `pyproject.toml`.
Users who installed `angr management` via either git or pip could then install these extra dependencies as needed via something like `pip install angrmanagement[docker]`, which could install `archr` and `keystone-engine` for docker image target support.
The pyinstaller build would then need to be built either without these options, preventing users from using these default plugins at all, or with all of these options, allowing users to use all of the default plugins.

zwimer:
@ltfish @rhelmot Thoughts?

rhelmot:
interaction view can probably be axed entirely. it was a cute idea from a hackathon long past, probably should have been a plugin.

zwimer:
What about other soft dependency things? This is an _incomplete_ list but a `grep -r ImportError -B 2` gives the snippet below. From this I see `bintrace`, `archr`, `keystone`, `angr.angrdb.AngrDB`, `nclib`, `slacrs`, `argument_resolver`, `qtpy`, `binsync`, `qtconsole`, `PySide6`, `sqlalchemy`, `pyqodeng` and `xdg`; I realize some of these are default included, but others are not.

For example, docker targets require  at least`archr` and `keystrone-engine`. We can either make these hard dependencies, remove this functionality and make it a non-default plugin, or make this a default plugin that requires extras installed via `pip install angrmanagement[docker]` where the pyinstaller build has this by default. I imagine at least some of the other soft dependencies are for other features where the same question is applicable.

```
zwimer@Lotus ~/D/W/a/angr-management>
angrmanagement/ui/main_window.py-try:
angrmanagement/ui/main_window.py-    from angr.angrdb import AngrDB
angrmanagement/ui/main_window.py:except ImportError:
--
[...]

PR #845, opened 2022-12-13
Do not show docker target action in file menu unless archr exists

zwimer:
Fixes: https://github.com/angr/angr-management/issues/835


angrmanagement/ui/views/interaction_view.py
```diff
@@ -295,6 +295,7 @@ def _start_interaction(self):
             req_msg = 'To use this feature you need to install the following:\n\n\t' + '\n\t'.join(is_missing)
             req_msg += '\n\nInstall them to enable this functionality.'
             req_msg += '\nRelaunch angr-management after install.'
+            req_msg += '\nNote: This cannot be done for pyinstaller builds.'
```
twizmwazin, line 298: I don't like this way of handling this. If we are in a frozen build and a feature is not available, we should just not show that feature. If for some reason we must show that the feature is there or disabled, the message should be more along the lines of "This feature is not available in this version of angr management".

twizmwazin (changes requested):

angrmanagement/ui/views/interaction_view.py
zwimer: I agree; that's why I made: https://github.com/angr/angr-management/issues/844
I'd like this soft-dependency stuff gone. But that's for a different PR.

twizmwazin: The interaction view might be able to go away soon, but archr definitely won't. If we can improve test quality and get it to work cross-platform it might be a hard dependency, but that's not realistically going to happen any time soon. Checking if it is a pyinstaller build should be easy enough and we can just add a switch to the error message. If `getattr(sys, "frozen", None)` returns something other than `None`, then we know we're in pyinstaller.

angrmanagement/ui/views/interaction_view.py
```diff
@@ -292,9 +293,12 @@ def _start_interaction(self):
...
         if len(is_missing) > 0:
-            req_msg = 'To use this feature you need to install the following:\n\n\t' + '\n\t'.join(is_missing)
-            req_msg += '\n\nInstall them to enable this functionality.'
-            req_msg += '\nRelaunch angr-management after install.'
+            if getattr(sys, "frozen", None) is None:
+                req_msg = 'To use this feature you need to install the following:\n\n\t' + '\n\t'.join(is_missing)
+                req_msg += '\n\nInstall them to enable this functionality.'
+                req_msg += '\nRelaunch angr-management after install.'
+            else:
+                req_msg = "This feature is not available on this build of angr management"
```
twizmwazin, line 301: `s/on/in`?

twizmwazin (approved):

PR #848, opened 2022-12-13
Remove deprecated QT checks and uneeded dep checks in __main__

twizmwazin:
This patch removes a bunch of dependency checks that should be enforced via pip. It also removes code to enable high dpi features that were needed with qt 5 but have been default since qt 6.

### DIFF
--- a/angrmanagement/__main__.py
+++ b/angrmanagement/__main__.py
@@ -7,7 +7,6 @@ import ctypes
 import threading
 import time
 import warnings
-import platform
 import signal
 from typing import Optional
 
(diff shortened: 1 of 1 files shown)


angrmanagement/__main__.py
```diff
@@ -97,53 +38,16 @@ def set_windows_event_loop_policy():
...
-    # workaround for https://bugreports.qt.io/browse/QTBUG-87014
-    # this is because the latest PySide6 (5.15.2) does not include this fix
-    v, _, _ = platform.mac_ver()
-    vs = v.split(".")
-    if len(vs) >= 2:
-        major, minor = [int(x) for x in vs[:2]]
-    else:
-        return
-    if major >= 11 or major == 10 and minor == 16:
-        os.environ['QT_MAC_WANTS_LAYER'] = '1'
```
mborgerson, line 110: Can you confirm this QT_MAC_WANTS_LAYER config is no longer needed

angrmanagement/__main__.py
```diff
@@ -97,53 +38,16 @@ def set_windows_event_loop_policy():
...
     from .config import FONT_LOCATION, IMG_LOCATION, Conf
 
-    # Enable High-DPI support
-    # https://stackoverflow.com/questions/35714837/how-to-get-sharp-ui-on-high-dpi-with-qt-5-6
-    if ("QT_DEVICE_PIXEL_RATIO" not in os.environ
-        and "QT_AUTO_SCREEN_SCALE_FACTOR" not in os.environ
-        and "QT_SCALE_FACTOR" not in os.environ
-        and "QT_SCREEN_SCALE_FACTORS" not in os.environ
-        ):
-        QCoreApplication.setAttribute(Qt.AA_EnableHighDpiScaling)
```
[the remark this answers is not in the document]
twizmwazin: First bullet under here says that it is no longer needed as high dpi is always enabled and not disableable. https://doc.qt.io/qtforpython/porting_from2.html#class-function-deprecations

angrmanagement/__main__.py
twizmwazin: The linked bug report suggests that it is fixed in newer versions of qt5 and qt6, and the pyinstaller bundle seems to start just fine on macos 12 with this patch.

PR #856, opened 2022-12-14
Remove binsync from main source tree

twizmwazin:
Binsync exists in its own repo


mahaloz:
@twizmwazin I've thought more on this about the two different repos, with BinSync changing frequently. How would you feel about me changing the BinSync side code so that the angr-management plugin is built-in to BinSync. Then we just place a single file (well two including the `plugin.toml`), that looks like this:

```
try:
    from binsync.angr import entry_point
except ImportError:
    print("BinSync is not installed!")
    return

entry_point(*args, **kwargs)
```

I took this idea from @zwimer suggestions to BinSYnc.

rhelmot:
or we just use the pattern I designed where you can symlink binsync into ~/.local/share/angr-management/plugins... or we implement the thing kevin wanted where you can register plugins through setuptools hooks...

mahaloz:
> or we just use the pattern I designed where you can symlink binsync into ~/.local/share/angr-management/plugins... or we implement the thing kevin wanted where you can register plugins through setuptools hooks...

@rhelmot  we currently have our installer in BinSync do exactly that: we symlink the plugin code into our angrmanagement install. It's a huge pain in the ass when you are changing branches frequently on AM. Maybe this will not be an issue now, since there will not be a code overwrite in the commit log.

twizmwazin:
@mahaloz You shouldn't ever be editing the site packages of angr management. angr management can load plugins from a directory outside of site packages, and this is currently the best and intended way of loading plugins that aren't shipped with angr management. This method also works with pyinstaller builds and allows you to update angr management without deleting plugins.

PR #862, opened 2022-12-22
Format + improve setup.cfg

zwimer:


setup.cfg
```diff
@@ -1,48 +1,48 @@
 [metadata]
-name = angr-management
+name = angr_management
```
ltfish, line 2: Why are we renaming "angr-management" here?

rhelmot: while technically, this change doesn't matter (it is part of some PEP to normalize - and _ characters), I would prefer to keep the name the same as the repository name

zwimer: @rhelmot Are you saying you prefer "angr-management" or "angr_management"? I left it as is because the tool I used (a pre-commit hook I have added to the pre-commit PR) formatted it as such, but I could just not use this tool / remove this hook.

I'd be ok with just periodically running this hook manually from time to time to keep `setup.cfg` clean-ish, instead of having it be a pre-commit hook.

rhelmot: prefer angr-management, same as the repository name

twizmwazin (approved):

PR #871, opened 2023-01-11
Fix graph to display nodes with the same address (possible in AIL)

mahaloz:


mahaloz:
ummm @twizmwazin is CI broken? Something about the bundle build failing?

twizmwazin:
@mahaloz Should work if you rebase

PR #892, opened 2023-02-14
Use CFB on_object_added callback for live feature map updates

mborgerson:
Brings back live CFB updates. There is room for improvement, but this at least restores the previous level functionality with better performance.

Needs angr/angr#3806

Pending:
- [x] Fix races

Demo:

https://user-images.githubusercontent.com/8210/218686417-83a97f90-545b-4c55-88b2-511c4bab2e79.mp4


xxr0ss:
Super cool!🥳

PR #901, opened 2023-02-17
Update python for pyinstaller to 3.11

twizmwazin:


twizmwazin:
At this point most functionality seems to work, with the exception of the console view. I'm unsure if this is strictly due to the exclusion of debugpy (seemingly soft-required in ipykernel). Builds are available in CI if someone else wants to poke around at these builds and find problems.

ltfish:
YAY

PR #963, opened 2023-03-28
Remove workspace member from instance

twizmwazin:
This PR removes the workspace member from instance. This is a major step in separating the data and logic portions of angr management from the UI portions. Instance still retains some UI code for job progress display, but it does so without a direct reference to workspace, and I plan to move this code soon with some sort of JobExecutor abstraction.


twizmwazin:
Tests pass locally, seems to be some sort of weird CI/nose issue, investigating

PR #964, opened 2023-03-28
Allow double clicking on log rows to expand to show full log message

### LINKED ISSUE #947: Clicking on a truncated log message should show the user the full log message
### Description

When an exception is raised and reported in the log view, it is currently not possible to view past the first line of the exception output except by copying the message and pasting it in another program. It would be great for the message to either be expanded in the log view, or create a pop-out dialog showing the full text.

### Alternatives

_No response_

### Additional context

_No response_

twizmwazin:
Fixes #947, allows the user to expand a log entry to view its full contents by double clicking. Double clicking again will collapse the row to standard height. One minor issue: I chose to keep the text aligned to the top and left, but the icon is center aligned.


ltfish:
Good job!

Issue #974, opened 2023-04-06
console fails on latest nightly


psifertex:
### Description

On latest nightly, 9.2.46.dev0, on macOS Ventura 13.3 as well as Windows 22H2 (probably linux but I haven't tried that) the python console is unusable. You can enter any command (successful or not) and it simply stops being interactable afterward. The prompt never shows up again. If you close/reopen the console you can use it again and even have the same state but that's... painful.

<img width="742" alt="Screenshot 2023-04-06 at 2 01 15 PM" src="https://user-images.githubusercontent.com/140215/230459017-69d3a619-3549-440f-8750-babf93443a03.png">

### Steps to reproduce the bug

Run any command in the python console in the console. 

### Environment

Just using the nightly release of angr-management, no other environment.

xxr0ss:
Just tried latest angr-management on Windows 11, python 3.11. Console works well, no stops occur

![image](https://user-images.githubusercontent.com/31790712/230463904-73173e53-f9b8-4f7d-a422-dfd4688a1ce1.png)

twizmwazin:
It is caused by QtConsole 5.4.2. 5.4.0, 5.4.1 both work.

twizmwazin:
Should be fixed on master now, you can try out the builds from this pipeline run until the next nightly build is published: https://github.com/angr/angr-management/actions/runs/4632541869?pr=975

psifertex:
Thanks!

At least ya'll haven't had to deal with the messes in the 6.x line. Keep letting those shake out, it's been... rough.

![7hb72h](https://user-images.githubusercontent.com/140215/230489815-af63faaa-6314-43f9-8a90-04fa552661b2.jpg)

ltfish:
> Thanks!
> 
> At least ya'll haven't had to deal with the messes in the 6.x line. Keep letting those shake out, it's been... rough.
> 
> I guess to be fair though QtConsole isn't the mainline Qt folks fault so I was mis-attributing.

Thanks for reporting!

We did have to track Qt bugs, plus we tracked PySide bugs and bugs in all surrounding packages… it’s not fun, obviously.

PR #975, opened 2023-04-06
Pin qtconsole to 5.4.1, fixes #947

### LINKED ISSUE #947: Clicking on a truncated log message should show the user the full log message

twizmwazin:


twizmwazin:
Whoops, that should have been #974

PR #979, opened 2023-04-14
Remove dockerfile

twizmwazin:
angr management runs fine without docker, and as a UI-oriented tool, docker isn't a good option for containerization.


ltfish:
I believe the docker file existed so that we could run angr management on less tested platforms on only a short notice. They were extremely useful during hackathons. Can we move the dockerfile and the two scripts into a `legacy` folder in case we need them again?

twizmwazin:
I'll move it under the packaging directory where it can grow old.

PR #980, opened 2023-04-18
Fix precise diff from BaseView change commit

mahaloz:


mahaloz:
For reference CI has been failing since https://github.com/angr/angr-management/commit/2cd10c19a6f82b9dfd7a34db2a772982bee2d4f4

PR #981, opened 2023-04-18
Re-add the BinSync plugin (as a stub)

mahaloz:
## Why
Previously, BinSync was removed from angr-management since the code was too large and changed too frequently, which resulted in PR changes to both the BinSync repo and this one. [Recently](https://github.com/binsync/binsync/commit/3ca7530595614f8d0d6651ac124224f2327179f6), BinSync has transitioned all plugins to be inside the `binsync` library, which has made the code needed for decompilers just a stub.

I'm re-adding BinSync since this code in angr-management should rarely change, and it's only two files (one is the description TOML).

### DIFF
--- a/angrmanagement/plugins/angr_binsync/__init__.py
+++ b/angrmanagement/plugins/angr_binsync/__init__.py
@@ -1,9 +1,3 @@
-import logging
+from binsync.decompilers.angr.plugin import BinSyncPlugin
 
-try:
-    # we need to import the plugin to register it
-    from binsync.decompilers.angr.plugin import BinSyncPlugin  # noqa: F401
-except ImportError:
-    logging.getLogger(__name__).error(
-        "[!] BinSync is not installed, please `pip install binsync` for THIS python interpreter"
-    )
+__all__ = ["BinSyncPlugin"]
--- a/angrmanagement/plugins/angr_binsync/__init__.py
+++ b/angrmanagement/plugins/angr_binsync/__init__.py
@@ -1,0 +1,4 @@
+try:
+    from binsync.decompilers.angr import *
+except ImportError:
+    print("[!] BinSync is not installed, please `pip install binsync` for THIS python interpreter")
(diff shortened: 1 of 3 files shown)


twizmwazin:
I am against re-adding binsync as a stub to the main repo. Unless we want to sell binsync as a core feature of angr management, the goal is to remove all non-core plugins from the repo itself and encourage users to install plugins separately.

ltfish:
> Unless we want to sell binsync as a core feature of angr management

I think that's exactly what we want (BinSync evolved into its current form from an angr management plugin).

twizmwazin:
In that case we should make this compatible with the pyinstaller builds before merging, since this implementation assumes a dev environment.

ltfish:
> add BinSync to the required packages of angr-management

What does that entail? Shipping BinSync with angr management every week in our weekly builds?

mahaloz:
> What does that entail?

@ltfish It would just mean adding `binsync >= 3.9.0` to the requirements. This also means it makes its way into whatever virtual environment is created from PyInstaller. However, I really do think merging it just like this is fine. It doesn't need to be available to PyInstaller builds right now unless you think differently.

twizmwazin:
The pyinstaller builds are what I believe most angr management users are using. It's how we ship builds for grants, and it's where I tend to hear most of the user feedback coming from. It doesn't make any sense to me that we would want binsync to be a default plugin for pip users but non-existent for pyinstaller users.

mahaloz:
@twizmwazin I got it to build on my machine, but it fails to grab my local Python...
```
Cannot find path @executable_path/../../../../Python3 (needed by /Users/mahaloz/.virtualenvs/angr_39/bin/python)
```

So maybe it doesn't work. I mean, the CI fails to bundle; I just don't know why on MacOS and Windows only. Advice?

ltfish:
@twizmwazin Can you please take a look? We are preparing for a demo.

twizmwazin:
After a discussion on what the next steps we want to take for plugin support are, we've decided against this. We will be improving support for plugins with third party dependencies soon.

zardus:
Hey guys, without this PR, angr-management has no native collaboration support. Can we re-have this discussion please?

PR #1005, opened 2023-05-16
Move pyinstaller build requirements into pyinstaller extra, remove bundle.py

twizmwazin:


ltfish:
Can we add to the README or angr doc on how to bundle angr management locally? It would be extremely helpful during hackathons.

PR #1042, opened 2023-07-13
Not the xrefs you want, but the xrefs you deserve

Cl4sm:


angrmanagement/ui/widgets/qccode_edit.py
```diff
@@ -543,6 +574,9 @@ def _initialize_context_menus(self):
         self.action_rename_node = QAction("Rename...", self)
         self.action_rename_node.triggered.connect(self.rename_node)
         self.action_rename_node.setShortcut(QKeySequence("N"))
+        self.action_xref = QAction("XREFS...", self)
```
ltfish, line 577: Can we change it to `"XRefs..."`?

PR #1045, opened 2023-07-17
Add missing C keywords for highlighting.

adamdoupe:
`extern` wasn't highlighted, so I added missing C language keywords from here:
https://en.wikibooks.org/wiki/C_Programming/Language_Reference

Also sorted the highlighting rules array (within the functions).


ltfish:
Thanks!

PR #1047, opened 2023-07-19
Pseudocode: highlight based on AST instead of regular expressions.

adamdoupe:
Change how pseudocode is highlighted by using the AST information instead of hard-coded regular expressions.

Also add ability for themes to change color of types as well as variables.


ltfish:
Fix linting and I'll click merge!

PR #1097, opened 2023-09-24
FunctionDialog: Syntax highlight function prototype

mborgerson:
Dialog looks a little nicer now.

Needs updated pyqodeng

image


mborgerson:
@ltfish Would you please push a pyqodeng update

PR #1103, opened 2023-10-09
Update bundled python to 3.12

twizmwazin:


twizmwazin:
Looks like we're still waiting on pyside: https://bugreports.qt.io/browse/PYSIDE-2230

ltfish:
I'm glad that you opened this issue! Didn't notice before.

twizmwazin:
Looks like pyside6 6.6.0 is supporting 3.12 now, we need to rebuild pyside6-qtads to support it, right now it is pinned to 6.5.2.

PR #1114, opened 2023-11-13
Use full pyside6

twizmwazin:


twizmwazin:
In the process of testing PyInstaller builds, do not merge yet

PR #1118, opened 2023-11-15
Use UDBZ format for macos disk image

twizmwazin:
This'll hopefully get the macOS bundle size more under control


twizmwazin:
This does shrink the dmg by 18%, but it adds 8 minutes to the build time to do so, I'm not sure that's a worthwhile tradeoff.

ltfish:
> This does shrink the dmg by 18%, but it adds 8 minutes to the build time to do so, I'm not sure that's a worthwhile tradeoff.

I think it's worthwhile.

twizmwazin:
I've switched over to using zips for macos.

PR #1129, opened 2023-11-17
Add Catppuccin Mocha theme

agnosticlines:
Just added a theme I personally use across most of my tools, here's some screenshots

D3MUrD6Lel

xu2UzecoaS

I think it looks pretty decent, still working on it, I want to fix the buttons looking slightly too dark that might imply they're disabled but it's about 90% of the way there.


ltfish:
Very cool!

PR #1164, opened 2024-02-02
Automatically load theme on selection instead of using load button

### LINKED ISSUE #1163: Remove "Load" button when selecting a theme
### Description

Having a load button separate from the save button is unintuitive. I thought initially that theming was actually broken entirely. A user should only have to select their theme and not require additional steps.

### Alternatives

_No response_

### Additional context

_No response_

twizmwazin:
Fixes #1163 

Preview:

https://github.com/angr/angr-management/assets/1341797/b09c1cd0-84c9-4ba8-96da-98643ec37f18


mborgerson:
Much better!

PR #1220, opened 2024-03-21
Use from __future__ import annotations

twizmwazin:
This tells python to treat type annotations as strings instead of evaluating them on import. In practice, this means two things: we don't have to quote our type annotations, and we can use syntax from newer python versions that is more concise (ie, `A | B` instead of `Union[A, B]` or `A | None` instead of `Optional[A]`, `list[A]` instead of `from typing import List; List[A]`). This behavior will become standard in a future version as well. Ruff will automatically add this line to files in new commits to maintain consistency.

Related PEP: https://peps.python.org/pep-0563/


ltfish:
What’s the minimum Python version that supports this?

twizmwazin:
>What’s the minimum Python version that supports this?

3.7

PR #1226, opened 2024-03-22
Automatically add a bunch of types

twizmwazin:
This branch uses the [autotype](https://github.com/JelleZijlstra/autotyping) project to add a lot of type annotations to angr management. The goal here is to improve type coverage, and to avoid spending lots of time annotating the "obvious" stuff, but still ensuring that mypy has enough information to usefully report errors. This approach used with this tool is fairly conservative, and while it will have some errors that will need to be manually corrected, overall it should be an improvement.


xxr0ss:
Hi, I think `None` returns for class constructor `__init__` are the "obvious" stuff that should be manually corrected

twizmwazin:
What do you mean? In python constructors always return None. See here: https://docs.python.org/3/reference/datamodel.html#object.__init__

xxr0ss:
I mean maybe we don't need a `-> None` annotation for `__init__`, while autotype seems to add such annotation for every `__init__`

twizmwazin:
I made that choice intentionally, autotype only adds what you tell it to. The goal is to type annotate everything in the end and run it through a static type checker in CI (mypy). While perhaps it could be inferred, I’ve seen enough errors in magic methods not returning the correct type that I think it’s justified.

xxr0ss:
Understood, thanks for your explanation!

PR #1230, opened 2024-03-26
Preserve symbolic links when creating macOS bundle

twizmwazin:
If this works, this should cut the macOS bundle size by nearly 50%...


twizmwazin:
Just kidding, try 77%...

PR #1236, opened 2024-03-30
Add Windows CI

twizmwazin:


twizmwazin:
As far as I can tell, the macOS failures are a qt bug :(

twizmwazin:
My plan is to merge the windows half since it works, and I'll open a separate PR for macOS.

PR #1237, opened 2024-04-01
Fix and unify argument order of __init__ of instance views.

ltfish:


ltfish:
Superseded by PR #1235.

PR #1240, opened 2024-04-05
Add macOS CI

twizmwazin:


twizmwazin:
Seems macOS still segfaults on 6.7, I was hoping this was just a classic pyside broken build that would magically fix itself in the next update. I suspect the issue has to do with QT not running on the main thread, macOS is picky about which threads GUI applications are using. The good news is there is a pytest [plugin](https://github.com/pytest-dev/pytest-qt) that fixes this! Bad news, it is a pytest plugin, and to use it means marrying our tests to pytest which we have done up to this point, and isn't compatible with the angr ci. So if we use it, it needs to be done in an optional way, at least for now.

PR #1280, opened 2024-07-16
pyinstaller: Add pyxdia data

mborgerson:


mborgerson:
Confirmed working on macOS arm64

PR #1357, opened 2024-11-26
Update binsync dependency

### LINKED ISSUE #1356: Depending on super old version of binsync for seemingly no reason
### Description

I checked the diff, looks like there are no API changes from v4.0.0 (currently used) to v5.2.0 (current latest), should be safe to update as far as I can tell 

_No response_

### Environment

_No response_

### Additional context

_No response_

spencerpogo:
Closes #1356

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -27,7 +27,7 @@ install_requires =
     tomlkit
     pyobjc-framework-Cocoa;platform_system == "Darwin"
     thefuzz[speedup]
-    binsync==4.0.0
+    binsync>=4.0.0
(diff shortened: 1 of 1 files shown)


twizmwazin:
I think I would prefer to keep with a pinned version rather than a >= requirement. A >= requirement is bound to cause an issue someday.

spencerpogo:
Agreed. Although this loses the information that we support 4.x, it is overall better

spencerpogo:
No need for range I think it's fine as it stands right now

PR #1403, opened 2025-02-21
CodeView: Fix xrefs by passing the correct variable manager.

ltfish:


ltfish:
PySide6 does not like to be type checked. Honestly there is very little I can do.

PR #1421, opened 2025-05-15
Nightly bundle: Install Rust and setuptools-rust.

ltfish:


rhelmot:
Corpus decompilation diffs can be found at https://github.com/angr/dec-snapshots/compare/master...angr/angr-management_1421

PR #1438, opened 2025-07-08
Remove interaction view

twizmwazin:
This is unused and is the second place besides the button that archr is used in angr management.


rhelmot:
Corpus decompilation diffs can be found at https://github.com/angr/dec-snapshots/compare/master...angr/angr-management_1438

rhelmot:
god bless. I was really hoping this shit I wrote for a CHESS hackathon would actually be useful but we seem to have figured out the answer to that :)

PR #1440, opened 2025-07-11
Generate an installer for Windows

mborgerson:
I was reminded how painfully slow unzipping release builds are, so I've finally gotten 'round to adding a nice installer via NSIS:

image

* Familiar interface
* License notice
* Configurable installation options
* Desktop icons
* Better compression ratio compared to current (326M vs 221M)
* Faster decompression
* Uninstaller

TODO:
- [ ] Fix version information
- [ ] Add tests


installer.nsi
```diff
@@ -0,0 +1,145 @@
+; angr-management NSIS installer builder script
+; ==============================================================================
+; Build with: makensis -V4 -DW64 -DVERSION=1.2.3.4 installer.nsi
+
+Unicode true
```
mborgerson, line 5: Move

PR #1450, opened 2025-07-21
Fix: Undefining and defining does not bring back the same CFG nodes.

ltfish:


rhelmot:
Corpus decompilation diffs can be found at https://github.com/angr/dec-snapshots/compare/master...angr/angr-management_1450

PR #1476, opened 2025-08-20
CodeView: Fix a race condition between setting current line and updating _last_function.

ltfish:
This bug causes the pseudo code view to stay at the previous line number after switching to a newly decompiled function.


rhelmot:
Corpus decompilation diffs can be found at https://github.com/angr/dec-snapshots/compare/master...angr/angr-management_1476

PR #1574, opened 2026-02-02
Fix UberEnginePcode being selected for VEX arches

twizmwazin:
This fixes the issue in https://github.com/angr/angr-management/pull/1573. After the archinfo change, the heuristic we were using to determine if an arch was a pcode arch was to see if the parameter was defined, but now that name is a function defined on the base class. So let's just check directly instead.


ltfish:
Closed in favor of #1573.

PR #1600, opened 2026-02-27
QDecompilationOptions: Fix the bug where Boolean values are displayed as strings or ints

tedanvosin:
Check `option.value_type` using `isinstance` and not `issubclass`. 
Since `bool` is a subclass of `int`, and all boolean decomp-options appear as integers, this change will fix that.


tedanvosin:
You are right, I missed that. Though the suggested fix also does not seem right as that will make `BoolPropertyItem` for options that have multiple options, such as structuring algorithms as their `value_type` is `type`.

This updated patch should fix the issue(hopefully).

PR #1619, opened 2026-03-25
Changed comment prefix and shortcut

### LINKED ISSUE #1583: DisassemblyView comment token is C++ style comment
### Description

Typical assembly comment token is `;`. The shortcut to make a comment in disassembly view is also <kbd>;</kbd>. IMO the default comment character should be `;` or the default keyboard shortcut should be closer to other apps with <kbd>Ctrl</kbd>+<kbd>/</kbd>. Both could be customizable.

N/A

### Environment

angr-management @ 5cf584e7d39b089963ba13c50eeaff9a7c60599c

### Additional context

_No response_

NikhilC2209:
This PR changes the COMMENT_PREFIX from `;` to `//` in the `Disassembly View` to match assembly-style conventions. Also added `Ctrl+/` as an additional shortcut for opening the comment dialog.

Closes #1583

### DIFF
--- a/angrmanagement/ui/views/disassembly_view.py
+++ b/angrmanagement/ui/views/disassembly_view.py
@@ -294,8 +294,11 @@ class DisassemblyView(SynchronizedFunctionView):
             # decompile
             self.decompile_current_function()
             return
-        elif (key == Qt.Key.Key_Semicolon or
-             key == Qt.Key.Key_Slash and event.modifiers() == Qt.KeyboardModifier.ControlModifier):
+        elif (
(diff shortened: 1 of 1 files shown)


NikhilC2209:
Okay, I reverted the change to the `COMMENT_PREFIX` but kept the shortcut intact as `Ctrl+/`.

ltfish:
My bad @NikhilC2209 I thought you were changing from `; ` to `// `. I think `; ` is a better choice. I apologize. Can you please drop your most recent commit?

PR #1620, opened 2026-03-26
RetypeNode: Use node.type to show original type for CVariable

tedanvosin:
- Right now when you try to modify the type of a variable, original type is never shown as `self._node_type` is passed from `qcode_edit.py:retype_node`. This is fixed by using `self._node.type` direclty.


ltfish:
Thanks!

PR #1631, opened 2026-04-06
QCodeEdit: Update prototype and retype function when retyping arguments

tedanvosin:
Right now only way to retype a function argument is by retyping the function, which does not seem very intuitive. 

This change fixes that by checking if the retyped variable is an argument and updating the function prototype with the updated argument type.


ltfish:
Merged. Thanks!

PR #1632, opened 2026-04-11
Update flirt_signatures to the latest version

bluesadi:
## Summary

Bumps the `flirt_signatures` submodule to the latest upstream revision. Upstream has been restructured into a Python package layout, so the actual `.sig` / `.meta` files now live one directory deeper, under `angrmanagement/resources/flirt_signatures/flirt_signatures/`.

Without this PR the default `flirt_signatures_root` still points at the old top-level directory and `angr.flirt.load_signatures` finds nothing.

## Changes

- Submodule `angrmanagement/resources/flirt_signatures` bumped to the latest commit.
- `angrmanagement/config/config_manager.py`: default of `flirt_signatures_root` changed from `./angrmanagement/resources/flirt_signatures/` to `./angrmanagement/resources/flirt_signatures/flirt_signatures/` so the nested layout is picked up in both dev checkouts and PyInstaller builds.


mborgerson:
Thanks!

PR #1637, opened 2026-04-16
RetypeNode: Add tests

tedanvosin:
Tests for RetypeNode Dialog.


tedanvosin:
Failing testcase fixed with #1638

PR #1642, opened 2026-04-20
Fix crash when editing a struct or union field

emmanuel-ferdman:
### PR Summary
In `edit_field`, `dialog.result` was being treated as a list of edited fields, but `CTypeEditor` never sets that attribute - `dialog.result` resolves to the inherited `QDialog.result()` bound method. So `dialog.result[0]` raises `TypeError: 'builtin_function_or_method' object is not subscriptable` when user edits a struct or union field. The preceding guard `if not dialog.result:` was also silently broken: bound methods are always truthy, so the Cancel early-return never fired either. Both references now use `dialog.main_result`, the attribute `CTypeEditor` actually populates when the user accepts the dialog.


rhelmot:
Looks good, thanks!

PR #1647, opened 2026-04-27
RenameNode: Move argument renaming code and update prototype when renaming arguments

tedanvosin:
- Move function argument renaming logic into unified_variable is not none branch, as arguments have unified_variables, so the older branch is never taken.
- Update arg_names in function prototype to match changes made while renaming arguments.

### DIFF
--- a/angrmanagement/ui/dialogs/rename_node.py
+++ b/angrmanagement/ui/dialogs/rename_node.py
@@ -194,14 +194,14 @@ class RenameNode(QDialog):
                 # function argument
                 elif self._node.unified_variable.is_function_argument:
                     workspace.plugins.handle_func_arg_renamed(
-                        code_kb.functions[self._node.codegen.cfunc.addr], 0, self._node.variable.name, node_name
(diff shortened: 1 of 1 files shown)


angrmanagement/ui/dialogs/rename_node.py
```diff
@@ -187,6 +186,18 @@ def _on_ok_clicked(self) -> None:
...
+                elif self._node.unified_variable.is_function_argument:
+                    workspace.plugins.handle_func_arg_renamed(
+                        code_kb.functions[self._node.codegen.cfunc.addr], 0, self._node.variable.name, node_name
+                    )
+                    arg_names = list(self._node.codegen.cfunc.functy.arg_names)
+                    for idx, arg in enumerate(self._node.codegen.cfunc.arg_list):
+                        if arg is self._node:
+                            arg_names[idx] = node_name
+                            break
+                    self._node.codegen.cfunc.functy.arg_names = tuple(arg_names)
```
mborgerson: Why set all `arg_names` here?

PR #1652, opened 2026-04-27
RenameNode: Update FunctionView when a Function is renamed

tedanvosin:
- Add a `on_function_updated` function to `Workspace` to refresh FunctionView.
- Call `Workspace.on_function_updated` when a function is renamed through a `CFunction` or a `CFunctionCall`

### DIFF
--- a/angrmanagement/ui/workspace.py
+++ b/angrmanagement/ui/workspace.py
@@ -274,6 +274,11 @@ class Workspace:
         if view is not None:
             view.reload()
 
+    def on_function_updated(self) -> None:
+        functions_view = self.view_manager.first_view_in_category("functions")
+        if functions_view is not None:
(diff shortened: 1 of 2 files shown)


angrmanagement/ui/workspace.py
```diff
@@ -274,6 +274,11 @@ def on_flirt_signatures_applied(self) -> None:
         if view is not None:
             view.reload()
 
+    def on_function_updated(self) -> None:
+        functions_view = self.view_manager.first_view_in_category("functions")
```
mborgerson, line 278: Eventually we'll migrate away from this pattern of workspace triggering individual view updates (views should subscribe to changes), but for now this is ok

mborgerson:
Can you fix the typecheck failures, then we'll merge

tedanvosin:
Fixed

PR #1653, opened 2026-04-27
InfoDock: Unselect operands when selecting BlockTreeNode

tedanvosin:
Resolves the issue where an operand is left selected when you select a Function Header after selecting an operand.


mborgerson:
Thanks!

PR #1658, opened 2026-04-29
MainWindow: Set MainWindow as parent of exit message

tedanvosin:
Resolves the issue where clicking outside the "Save database?" dialog on exit, the message box would go behind the angr-management window, making it appear unresponsive.


mborgerson:
Thanks!

PR #1659, opened 2026-04-29
Allow Loading from an archive

tedanvosin:
- Adds ArchiveLoaderDialog, which shows contents of an archive file with Name and size. 
- Selected file is extracted to a tmp directory, and the path is returned to AM for loading as a regular file.


mborgerson:
Please add tests for this new feature

mborgerson:
Considering that we may be analyzing malware in encrypted Zips, we should probably also support:
* Providing password for encrypted archives
* Extracting in-memory (so malware is never placed on the filesystem)

But these can come in a follow up.

PR #1661, opened 2026-05-04
QLinearDisassembly: Do not use cache when reloading

tedanvosin:


angrmanagement/ui/widgets/qlinear_viewer.py
```diff
@@ -109,7 +109,7 @@ def reload(self, old_infodock: InfoDock | None = None) -> None:  # pylint:disabl
         curr_offset = self._offset
         self.initialize()
         self._offset = None  # force a re-generation of objects
-        self.prepare_objects(curr_offset, start_line=self._start_line_in_object)
+        self.prepare_objects(curr_offset, start_line=self._start_line_in_object, use_cache=False)
```
mborgerson, line 112: Why parameterize cache use instead of just clearing the cache (which the above line indicates)?

PR #1662, opened 2026-05-04
QLinearDisassembly: Clear objects cache in initialize

tedanvosin:


angrmanagement/ui/widgets/qlinear_viewer.py
```diff
@@ -236,6 +236,13 @@ def initialize(self) -> None:
         if self.cfb.am_none:
             return
 
+        for obj in self.objects.values():
+            if obj.isVisible():
+                obj.setVisible(False)
+            obj.remove_children_from_scene()
+            self.scene.removeItem(obj)
+
+        self.objects.clear()
```
mborgerson, line 245: does `clear()` call the eviction handler for us already (which calls remove_children_from_scene and removeItem)?

PR #1687, opened 2026-06-18
QDisasmStatusBar: Use demangled function name

tedanvosin:
- Use demangled function in disasm status bar.
- Truncate long function names, to avoid the extra scroll bar.


mborgerson:
Thanks!

PR #1692, opened 2026-06-29
CI: brew install keystone on MacOS

tedanvosin:


twizmwazin:
Shouldn't we be using the version supplied from pypi?

rhelmot:
Version installed from pypi has not been changed in 6 years

PR #1716, opened 2026-08-11
Resolve referenced sibling pull requests in the test jobs

zardus:

The `Test` jobs in ci.yml and coverage.yml run their own `uv sync`, and `[tool.uv.sources]` pins archinfo, pyvex, cle, claripy, angr and bintrace to a git branch of master, so a change needing a sibling change was built without it. This extends the binaries-only scan added here to all six, in the two spellings `resolve_refs.py` accepts, taking the first reference to each while it is open.

The reference becomes `no-sources-package` plus `upgrade-package` in a uv configuration file under the runner temporary directory, so uv resolves that package from the pull request head, keeps the rest on master, and writes nothing into the checkout. angr/angr#6821 and angr/cle#738 now carry this same mechanism.

Validation: https://github.com/angr/angr-management/pull/1716#issuecomment-5259983416.

The `pyinstaller / Build on macos-15-intel` check is red for a reason unrelated to this change: building angr's Rust extension there fails in the `core_affinity2` crate with `error[E0308]: mismatched types` at `core_affinity2-0.16.1/src/lib.rs:885`, which `libafl 0.16.1` pulls in. That is angr/angr#6890. Every other check on this pull request passes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,6 +79,43 @@ jobs:
       - uses: ilammy/msvc-dev-cmd@0b201ec74fa43914dc39ae48a89fd1d8cb592756 # v1
         if: startsWith(runner.os, 'windows')
       - uses: astral-sh/setup-uv@85856786d1ce8acfbcc2f13a5f3fbd6b938f9f41 # v6
+      - name: Resolve referenced sibling pull requests
+        shell: bash
+        env:
+          GH_TOKEN: ${{ github.token }}
+          PR_BODY: ${{ github.event.pull_request.body }}
+        run: |
+          config=$RUNNER_TEMP/uv-sources.toml
+          packages=
+          requirements=
+          seen=
+          references=$(printf '%s' "$PR_BODY" |
+            grep -Eo '[A-Za-z0-9._-]+/[A-Za-z0-9._-]+#[0-9]+|github\.com/[A-Za-z0-9._-]+/[A-Za-z0-9._-]+/pull/[0-9]+' |
+            sed -E -e 's|^github\.com/||' -e 's|/pull/|#|' -e 's|[/#]| |g') || true
+          while read -r owner repo number; do
+            [ -n "$number" ] || continue
+            case " $seen " in *" $owner/$repo "*) continue ;; esac
+            seen="$seen $owner/$repo"
+            package=$(sed -n '/^\[tool\.uv\.sources\]/,/^\[/p' pyproject.toml |
+              grep -F -e "\"https://github.com/$owner/$repo\"" -e "\"https://github.com/$owner/$repo.git\"" |
+              sed -E 's/^ *([^ =]+).*/\1/' |
+              head -n 1) || true
+            [ -n "$package" ] || continue
+            state=$(gh api "repos/$owner/$repo/pulls/$number" --jq .state) || state=unavailable
+            if [ "$state" != open ]; then
+              echo "$owner/$repo#$number is $state, so it is not used"
+              continue
+            fi
+            echo "Installing $package from $owner/$repo#$number"
+            packages="$packages\"$package\","
+            requirements="$requirements\"$package @ git+https://github.com/$owner/$repo@refs/pull/$number/head\","
+          done <<<"$references"
+          if [ -z "$packages" ]; then
+            echo "No sibling pull request is referenced, so every sibling stays on master"
+            exit 0
+          fi
+          printf 'no-sources-package = [%s]\nupgrade-package = [%s]\n' "$packages" "$requirements" | tee "$config"
+          echo "UV_CONFIG_FILE=$config" >>"$GITHUB_ENV"
       - name: Sync dependencies
         run: uv sync -p 3.12
       - name: Run tests
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,10 +25,33 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v4
+      - name: Resolve the angr/binaries ref
+        id: binaries-ref
+        shell: bash
+        env:
+          GH_TOKEN: ${{ github.token }}
+          PR_BODY: ${{ github.event.pull_request.body }}
+        run: |
+          ref=master
+          number=$(printf '%s' "$PR_BODY" |
+            grep -Eo '(angr/binaries#|github\.com/angr/binaries/pull/)[0-9]+' |
+            head -n 1 |
+            grep -Eo '[0-9]+$') || true
+          if [ -n "$number" ]; then
+            state=$(gh api "repos/angr/binaries/pulls/$number" --jq .state) || state=unavailable
+            if [ "$state" = open ]; then
+              ref="refs/pull/$number/head"
+            else
+              echo "angr/binaries#$number is $state, so it is not used"
+            fi
+          fi
+          echo "Checking out angr/binaries at $ref"
+          echo "ref=$ref" >>"$GITHUB_OUTPUT"
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v4
         with:
           repository: angr/binaries
           path: binaries
+          ref: ${{ steps.binaries-ref.outputs.ref }}
       - name: Move binaries
         run: mv binaries ..
       - name: Setup Rust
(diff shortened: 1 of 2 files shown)


zardus:

Validation record for head `d060e080a06b96bf803cc6776e31830334b3e3d8` against baseline `1de552537627152c06a673c5f917083823fde1ec`. The head is unchanged: this is the mechanism the other two have now adopted, so nothing here needed to move. All three of angr/angr#6821, angr/cle#738 and angr/angr-management#1716 now use one mechanism, and the two shared steps are byte-identical across them.

- Mechanism: the reference becomes `no-sources-package` plus `upgrade-package` in a uv configuration file under `$RUNNER_TEMP`, and `UV_CONFIG_FILE` carries it to every later uv command in the job, so the override sits inside uv's own resolver instead of being applied on top of a finished sync. One resolution, one lock, and nothing to undo afterwards. Both settings are needed and neither works alone, re-checked on uv 0.12.1 against a throwaway project pinning archinfo to `branch = "master"`: with `no-sources-package` alone the pin is dropped and archinfo resolves from PyPI, `source = { registry = "https://pypi.org/simple" }`, not from the pull request; with `upgrade-package` alone resolution fails outright, ``Requirements contain conflicting URLs for package `archinfo` in all marker environments: - git+https://github.com/angr/archinfo.git@master - git+https://github.com/angr/archinfo@refs/pull/367/head``; with both, `source = { git = "https://github.com/angr/archinfo?rev=refs%2Fpull%2F367%2Fhead#a5f11be91c0c6f4483dfb9bf7e3517f73febb7a4" }`, that pull request's head. The alternatives remain unavailable: a `uv.toml` rejects a `sources` table outright, and `override-dependencies` is ignored for a package that already has a source.
- `UV_CONFIG_FILE` does not displace the project's own `[tool.uv]`: with the file set, `default-groups` from `pyproject.toml` still applied and the extra group was still installed, so the only thing the file changes is the package being overridden.
- End to end, unchanged from the run this record already covered: with the description naming angr/cle#738, every job that runs `uv sync` logged `Installing cle from angr/cle#738`, wrote `no-sources-package = ["cle",]` with `upgrade-package = ["cle @ git+https://github.com/angr/cle@refs/pull/738/head",]` to `$RUNNER_TEMP/uv-sources.toml`, and installed `cle==9.3.3.dev0 (from git+https://github.com/angr/cle@9ad26c0)`, that pull request's head, instead of cle master `45c6509`, while archinfo, pyvex, claripy and angr all came from their `master` branches and `uv run pytest` afterwards re-synced nothing: ci.yml [ubuntu-24.04-arm](https://github.com/angr/angr-management/actions/runs/31582417526/job/94068341048), [macOS](https://github.com/angr/angr-management/actions/runs/31582417526/job/94068340957), [Windows](https://github.com/angr/angr-management/actions/runs/31582417526/job/94068341045), coverage.yml [Build](https://github.com/angr/angr-management/actions/runs/31582416996/job/94068339154). Those same runs logged `Checking out angr/binaries at master`, so the two resolvers work independently.
- The worktree the sync leaves behind: `git diff --stat` empty, so no tracked file changed, and `git status --porcelain` reporting only `?? uv.lock`, which `uv sync` writes itself, plus `?? test_durations.json` in coverage.yml from the durations cache. The configuration file is written under `$RUNNER_TEMP` and never in the checkout. That check ran on `d1ac5bb8`, a commit carrying an extra step for this purpose that is not part of the change, everything else identical: ci.yml [ubuntu-24.04-arm](https://github.com/angr/angr-management/actions/runs/31580815289/job/94063207722), [macOS](https://github.com/angr/angr-management/actions/runs/31580815289/job/94063207670), [Windows](https://github.com/angr/angr-management/actions/runs/31580815289/job/94063207680), coverage.yml [Build](https://github.com/angr/angr-management/actions/runs/31580814859/job/94063205733)
- Resolution against the real dependency graph, locally: `uv lock` on this `pyproject.toml` with the file the step writes for the current description, which names both angr/angr#6821 and angr/cle#738, resolved the same 248 packages CI resolves in one pass, took angr from `refs/pull/6821/head` at `f892974` and cle from `refs/pull/738/head` at `b153602`, the current heads of those two, and left angr-data, archinfo, bintrace, claripy and pyvex on `master`. Overriding two siblings at once therefore resolves as cleanly as one.
- Resolver, the `run:` block taken verbatim out of both workflows and driven with `bash --noprofile --norc -eo pipefail`, the way a `shell: bash` step runs, against the real GitHub API. The two workflows carry byte-identical blocks:
  - `angr/cle#738` and the `github.com/angr/cle/pull/738` spelling each produce `no-sources-package = ["cle",]` and the matching `upgrade-package`; a body naming several siblings produces all of them
  - merged angr/cle#735 is reported closed and produces nothing; angr/binaries#176 and angr/angr-data#1, neither in `[tool.uv.sources]`, produce nothing and make no API call; prose with no reference, an empty body, which is what `${{ github.event.pull_request.body }}` evaluates to on a push and on a workflow_dispatch, and an unset `PR_BODY`, all write no file and leave `UV_CONFIG_FILE` unset
  - two references to one repository take the first
  - a body of ``angr/cle#738 $(touch …) `touch …` ; touch … && rm -rf / | id ${IFS} angr/cle#$(id -u) $(id)/$(id)#1 ../../etc#7 angr/cle;id#738`` resolves `refs/pull/738/head` and nothing else, executes none of it, and leaves the canary file it would have created absent; only digits reach the URL and the body reaches the step through `PR_BODY`
  - `angr/angr#6821` selects the angr entry and not angr-data, and `angr/angr-data#1` selects neither here, since the match is on the quoted URL and the trailing quote keeps one from matching the other
  - run against the exact live description text of all three pull requests, each resolves only its intended siblings; this description's own validation link matches the URL spelling but overrides nothing, angr-management not being in its own `[tool.uv.sources]`
- Binaries: unchanged, and byte-identical to the copies angr/angr#6821 and angr/cle#738 now carry. Driven the same way it gives `refs/pull/168/head` for the open angr/binaries#168, and `master` for merged angr/binaries#176, for a body naming only a sibling, and for an empty body. The earlier end-to-end runs stand: with the description naming angr/binaries#176 all five jobs logged `Checking out angr/binaries at refs/pull/176/head` and checked out `9141d93`, and with no reference all five logged `Checking out angr/binaries at master` and made no API call.
- Shared text: the `run:` script of `Resolve the angr/binaries ref` is sha256 `0297b9806c9d2f6ebd9a4c5d…` in all four workflows across the three repositories that carry it, and the `run:` script of `Resolve referenced sibling pull requests` is sha256 `b0e3646b08d1a133c4a60f0f…` in all four. cle's copy carries one extra YAML key, `working-directory: cle`, because its own checkout is a subdirectory of the workspace there; nothing inside the script differs.
- Lint: `actionlint 1.7.12` with `shellcheck 0.11.0` on both workflows reports the same five pre-existing SC2086 notes as the merge base and nothing new; `pre-commit run --all-files` in angr-management passes and leaves the tree unchanged.
- Not run: the workspace Python gate, since this changes no Python.

Everything on this head is green apart from the one job noted here. The one red check, coverage.yml `Report`, is unrelated: it fails at `Upload test coverage to Codecov` with `gpg: Can't check signature: No public key`, because the action cannot verify the Codecov CLI download. Every coverage.yml run in this repository's recent history fails there, including [master at `1de55253`](https://github.com/angr/angr-management/actions/runs/31305323422/job/93225802194), this branch's base. That job runs no uv command and is untouched here.

Caveats: a description edit does not retrigger CI and a re-run replays the original event payload, so a reference has to be in the description when the run starts. A GitHub API failure while checking that a referenced pull request is open leaves that sibling on master rather than failing the job. Only repositories `[tool.uv.sources]` takes from GitHub can be overridden, matched by URL, so a reference to anything else is ignored. Because the override goes through the resolver, a sibling head that genuinely conflicts with the rest of the graph fails the sync instead of being installed over an already-resolved environment; that is the intended trade for a coherent lock. pyinstaller-build.yml runs its own `uv sync` through nightly-build.sh and is left alone here.

zardus:

Closing this in favour of #1723, for the same reasons as angr/angr#6821.

This adds 122 lines across `ci.yml` and `coverage.yml`, because both the sibling-resolution block and the `angr/binaries` ref block are pasted into each of the two workflows. #1723 does the binaries half in 8 lines by calling the composite action in angr/ci-settings (angr/ci-settings#123), so the logic is defined once for the whole organisation.

The sibling-resolution half should not be here at all: pointing a Python sibling at a pull request branch belongs in `tool.uv.sources` in `pyproject.toml`, which behaves the same locally and in CI and needs no workflow code.

#1723 is currently red only because it references the action at `master`, where angr/ci-settings#123 has not landed; that clears on merge.

Apologies for the churn — I filed #1723 without checking this one first.

PR #1719, opened 2026-08-18
ci: Bump codecov-action to v7 so coverage uploads verify

zardus:

The Report job has been failing on every open pull request and on master. It dies at "Upload test coverage to Codecov" with `==> Could not verify signature. Please contact Codecov if problem continues`, and because the step sets `fail_ci_if_error: true` that becomes a red check on work that is otherwise fine.

The cause is the pin rather than anything in the tests. `codecov/codecov-action@5a10915` is release 5.5.1 from September 2025, and the CLI it downloads no longer verifies against Codecov's current signing key. angr hit the same wall and moved to 7.0.0 in June, where both upload steps succeed today; this brings angr-management to that same revision.

Only the two `uses:` lines change. Worth noting separately that this repository has no `.github/dependabot.yml`, which is why the pin went stale here while angr's was bumped automatically — that seemed like a policy question for maintainers rather than something to fold into this fix.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -129,14 +129,14 @@ jobs:
           junit=$(find . -type f -name 'junit.xml' | paste -sd, -)
           echo "junit=$junit" >> "$GITHUB_OUTPUT"
       - name: Upload test coverage to Codecov
-        uses: codecov/codecov-action@5a1091511ad55cbe89839c7260b706298ca349f7 # v5
+        uses: codecov/codecov-action@fb8b3582c8e4def4969c97caa2f19720cb33a72f # v7
         with:
           use_oidc: true
           fail_ci_if_error: true
           verbose: true
           files: ${{ steps.files.outputs.coverage }}
       - name: Upload test results to Codecov
-        uses: codecov/codecov-action@5a1091511ad55cbe89839c7260b706298ca349f7 # v5
+        uses: codecov/codecov-action@fb8b3582c8e4def4969c97caa2f19720cb33a72f # v7
         with:
           use_oidc: true
           fail_ci_if_error: true


zardus:

**Scope of the failure.** `Report` is FAILURE on open pull requests #1718, #1716, #1709, #1700, #1693 and #1690, and SUCCESS only on #1615, last updated 2026-05-07. `Test with coverage` also fails on `master` (runs at 2026-08-18T08:44 and 2026-08-17T19:58), so this is not caused by any pull request's diff.

**The failing step.** Job 95618142435, step "Upload test coverage to Codecov":

```
-> Downloading https://cli.codecov.io/latest/linux/codecov
-> Downloading https://cli.codecov.io/latest/linux/codecov.SHA256SUM
-> Downloading https://cli.codecov.io/latest/linux/codecov.SHA256SUM.sig
==> Could not verify signature. Please contact Codecov if problem continues
##[error]Process completed with exit code 1
```

**The pins.** `5a1091511ad55cbe89839c7260b706298ca349f7` is release 5.5.1, dated 2025-09-04, added here on 2025-12-22 and never bumped. `fb8b3582c8e4def4969c97caa2f19720cb33a72f` is release 7.0.0, dated 2026-06-07, which angr adopted on 2026-06-08.

**Evidence the target revision works.** On angr, runs 32158019696, 32157967254 and 32157858148 each report `Upload test coverage to Codecov = success` and `Upload test results to Codecov = success`.

**Local checks.** The file parses and both steps resolve to the new revision. `actionlint` with shellcheck reports 5 findings on the base and 5 on this head — all info-level SC2086 at lines 34, 70 and 116, none introduced here and none near the changed lines. `pre-commit run --files .github/workflows/coverage.yml` passes; its Python hooks skip, `check yaml` passes.

**What only CI can confirm.** The failure is inside a hosted action, so this change cannot be proven locally. This pull request's own `Report` check is the test.

**Not included.** This repository has no `.github/dependabot.yml`, which is why the pin went stale here while angr's `github-actions` ecosystem entry kept angr's current. Adding one is a policy choice for maintainers, so it is left out of this change.

PR #1723, opened 2026-08-24
ci: check out a referenced angr/binaries pull request

zardus:

The `angr/binaries` checkout carries no `ref`, so this workflow always builds against master. A pull request whose new test loads a fixture from an open `angr/binaries` pull request fails here no matter what its body references, which is the same gap `angr/cle` closed in angr/cle#738.

This resolves the ref with the shared action from angr/ci-settings and passes it to the existing checkout, so this workflow's own checkout options are untouched.

Blocked on angr/ci-settings#123, which adds the action; this references it at `master`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,9 +25,12 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v4
+      - id: binaries-ref
+        uses: angr/ci-settings/actions/binaries-ref@master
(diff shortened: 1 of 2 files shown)


zardus:

The failing jobs all stop in `Set up job`, before any checkout or test runs:

```
Can't find 'action.yml', 'action.yaml' or 'Dockerfile' for action 'angr/ci-settings/actions/binaries-ref@master'
```

That is the expected state while angr/ci-settings#123 is open: this references the action at `master`, where it does not exist yet. The red is not a property of this diff and there is nothing here to repair. Merging angr/ci-settings#123 and re-running the whole run clears it.

The workflow file was checked with `yaml.safe_load` before pushing, and the resolution logic itself was exercised against five pull-request body forms in angr/ci-settings#123.

